PR #130, opened 2026-07-24
feat: FTN wizard nodelist lookup with hub autofill

robbiew:
## Summary

The FTN Setup Wizard gains a **Node Lookup** row: after entering your FTN address, press Enter on it and the wizard downloads the network's nodelist, finds your node and its uplink hub, and autofills Hub Address / Hub Hostname / Hub BinkP Port with real values instead of the static per-network registry defaults. The values flow into `ftn.json` and `binkd.conf` through the existing save path.

- **FTS-5000 nodelist parser + lookup** (`internal/ftn/nodelist.go`): resolves each entry's full address from Zone/Region/Host/Hub context; `Lookup` finds your entry and its uplink (nearest enclosing Hub → net Host → Zone), resolving BinkP hostname/port from `INA:`/`IBN:` flags with an `f<node>.n<net>.z<zone>.<dns_suffix>` fallback (makes FidoNet work despite sparse INA flags).
- **Unlisted-node inference**: brand-new nodes not yet in the weekly nodelist still get their hub inferred from the net's Host/Hub lines, with a clear "not listed yet" note.
- **Zip-aware downloader** (`internal/ftn/nodelist_fetch.go`): magic-byte zip detection, nodelist-member scoring, nested-zip rejection, 16MB caps at both download and member layers.
- **Curated registry URLs**: `nodelist_url` added to `registry.json` for FidoNet z1-4 (stable `latest` alias), fsxNet, AmigaNet, SFNet, Micronet, Zer0net, HobbyNet — each verified live. Stale or heuristically unpickable sources deliberately excluded so the wizard never autofills dead hubs.
- **`.z##` member support + legacy nodelists** (follow-up commits): the zip member scorer now recognizes the FTN `.z##` naming convention (these are single-nested zip-compressed nodelists, unwrapped exactly one level with size caps at every layer; double nesting still rejected), and per policy decision, dated-but-valid nodelists from slow-moving networks are included — adding WWIVnet, DoRENET, VKRadio, and tqwNet (Agoranet remains out: FTP-only distribution).
- **Wizard UX**: explicit lookup row with per-state text (no nodelist / enter address first / downloading / found "BBS, Location (Sysop)" / inferred / failed), ESC-cancellable download, cached nodelist for re-lookup without re-download, stale confirmation cleared on address edit. All failure paths leave manual entry fully usable.

## Test plan

- 20+ new unit tests: parser fixtures (keywords, flags, CRLF/EOF, malformed lines), lookup inference table (hub/host/zone walk, downed-hub routing, nearest-hub, DNS fallback, error paths), zip extraction (member scoring, largest fallback, nested rejection), httptest-based download tests.
- TUI tests: row state text, autofill overwrite, failure-leaves-hub-untouched, cached synchronous apply, late-result drop after ESC, network-switch and address-edit state clearing.
- `go test ./...`, `go test -race` on both touched packages, `gofmt`, `go vet` all clean.
- Final review verified the FidoNet URL end-to-end: live download parsed through `ParseNodelist`+`Lookup` with correct uplinks across all four zones.

## Summary by CodeRabbit

* **New Features**
  * Added an FTN **Node Lookup** step to the setup wizard.
  * When a network nodelist URL is available, the wizard downloads it after entering your FTN address, identifies your node, and auto-fills hub address/hostname and the BinkP port.
  * Displays clear status for loading, missing URLs, validation issues, listed vs inferred hubs, and lookup errors, with the ability to cancel the download.

* **Bug Fixes**
  * Changing networks or addresses now clears stale lookup state; manual hub edits are preserved.

* **Documentation**
  * Updated the configuration docs with the Node Lookup flow, manual-entry steps, and skip scenarios.


robbiew:
Addressed the review feedback in three commits (b3f97eb, 3f7cca5, 92f9bbf):

- **HTTPS nodelist sources**: Micronet switched to HTTPS (byte-identical content, valid cert). SFNet stays HTTP — its cert doesn't cover the wiki subdomain (no SAN match), and a working HTTP nodelist beats none; autofilled values remain visible for sysop review before saving.
- **Stale hub fields after address edit**: new `hubAutofilled` flag — editing your address now restores lookup-autofilled hub fields to registry defaults, while manually-entered hub values are never clobbered.
- **ESC now truly cancels**: `fetchFTNNodelist` runs under a cancelable context invoked on ESC and on network switch.
- **Stale/cross-network results**: results are stamped with URL + a `lookupGeneration` counter (same pattern as the V3Net registry browser's `requestID`) and dropped on mismatch — covers both the network-switch race and cancel-then-retry against the same URL.
- **fsxNet registry test**: now fails if the fsxNet entry is missing entirely.

PR #174, opened 2026-08-04
fix(wfc): security hardening from WFC audit + expanded sysop guide

robbiew:
## Summary

Security audit of the WFC console and admin channel surfaced five findings; this PR addresses all of them and expands the sysop guide.

### Hardening
- **Terminal escape injection**: caller handles are now stripped of control bytes before the WFC TUI renders them (node table, event feed, details view), and `validateHandle` rejects control characters at registration. A hostile handle could previously replay escape sequences (title changes, screen clears) into the sysop's terminal.
- **Live re-authorization**: open wfc-admin sessions re-check `authorizeAdmin` every 30s and disconnect when it stops holding (WFC access disabled, user demoted or banned in the running daemon). Key removals via `ue`/`helper` still require a restart (daemon reads `users.json` at startup); documented.
- **Refresh rate-limit**: `CommandRefresh` no longer forces a snapshot rebuild more often than the configured tick interval, closing a CPU/log-flood nuisance from a compromised admin key.
- **Key-scan visibility**: unknown public-key offers are logged at debug level with their SHA256 fingerprint.

### Docs
- New "Console functions" reference (header, node table, statuses, details, event feed, refresh/reconnect)
- Key generation → transfer → registration walkthrough, and a troubleshooting section for the auth failure modes
- Security model updated: re-authorization window, revocation restart caveat, disclosure scope at `coSysOpLevel` (including invisible sessions — kept visible to admins by design), display sanitization

## Test plan
- New tests: escape-injection through all three views, `sanitizeTerminal` unit cases, `validateHandle` control-char rejection, refresh rate-limiting, re-auth watcher kick/shutdown
- `go test -race` on `cmd/vision3`, `internal/admin`, `internal/wfcui`; full `go test ./...` green; gofmt/vet clean

Note: sanitization is client-side — remote sysops need a rebuilt `wfc` binary to pick it up.

## Summary by CodeRabbit

* **Security Enhancements**
  * Administration sessions now disconnect when authorization is revoked.
  * Terminal control characters are removed from displayed node and event data.
  * Invalid or unsafe handles are rejected during account setup.

* **Performance**
  * Rapid refresh requests are rate-limited to reduce unnecessary updates.

* **Documentation**
  * Expanded the WFC console guide with setup, usage, security, and troubleshooting instructions.

* **Tests**
  * Added coverage for authorization changes, refresh rate limiting, input validation, and terminal-output sanitization.

### DIFF
--- a/cmd/vision3/wfc_admin.go
+++ b/cmd/vision3/wfc_admin.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"strings"
 	"time"
 
 	"github.com/gliderlabs/ssh"
@@ -34,6 +35,11 @@ var wfcEnabled func() bool
 // wfcAdminHandleKey is the context key used to stash the admin handle during
 // public-key authentication so wfcAdminSubsystem can re-verify it.
 type wfcAdminHandleKey struct{}
+
+// wfcAdminPubKey is the context key used to stash the marshaled public key that
+// authenticated the session, so authorization can be re-verified against the
+// key itself — not just the account — for the life of the session.
+type wfcAdminPubKey struct{}
 
 // wfcPublicKeyHandler is the SSH-level public-key auth handler for admin clients.
 // If the key is registered to a BBS user with sufficient access level, the
@@ -68,6 +74,7 @@ func wfcPublicKeyHandler(ctx ssh.Context, key ssh.PublicKey) bool {
 		return false
 	}
 	ctx.SetValue(wfcAdminHandleKey{}, u.Handle)
+	ctx.SetValue(wfcAdminPubKey{}, key.Marshal())
 	slog.Info("wfc-admin: public key accepted", "user", u.Handle, "addr", ctx.RemoteAddr())
 	return true
 }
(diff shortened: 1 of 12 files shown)


robbiew:
Both Copilot findings were valid and are fixed in ab14bd4.

**C1 controls (4 comments).** Correct gap — `sanitizeTerminal` and `validateHandle` only covered C0/DEL, leaving U+0080–U+009F, where U+009B is the 8-bit CSI. Both now reject the C1 range via a shared `isTerminalControl` helper, with test cases for U+009B (8-bit CSI) and U+0080 in the sanitizer, the three view tests, and handle validation.

Worth noting why this is safe here: CP437 input decodes its high bytes to printable codepoints (0x82 → é, 0x9B → ¢, 0x80 → Ç), all outside U+0080–U+009F, so rejecting the C1 rune range cannot break legitimate CP437 handles. Verified before making the change.

**lastTick monotonicity.** Also correct, and a real ordering bug: `tick()` captures `now` before contending for `tickMu`, so a refresh-forced tick can apply an older timestamp after a later periodic tick has already landed — moving `lastTick` backwards and widening the refresh window. `lastTick` now only advances. Added `TestLastTickIsMonotonic`, which fails against the previous code.

Full suite green, `-race` clean on the four touched packages, gofmt/vet clean.

robbiew:
All five CodeRabbit findings were valid and are fixed in cea6ba7. Each fix was driven by a test that fails against the previous commit.

**1. Session re-auth ignored the key (major).** Correct, and worse than described. `authorizeAdmin` looked the account up with `GetUser`, which does **not** skip soft-deleted users, while the key path (`FindByAuthorizedKey`) does -- so an account *deleted* mid-session kept its console alive, and deletion is an in-BBS action that mutates live memory. New `authorizeAdminKey(handle, keyBytes)` re-runs the full initial-auth predicate against the stashed key, and the watcher now uses it. `TestAuthorizeAdminKey` covers key removal, soft-delete, demotion, WFC disabled, and a key moved to a different account.

**2. Concurrent refresh duplicated work (major).** Confirmed: 16 simultaneous `CommandRefresh` calls produced **12** registry polls, because the freshness check ran outside `tickMu`. Extracted `refreshIfDue`, which takes `tickMu`, then reads the clock and re-checks `lastTick` inside it. `TestConcurrentRefreshPollsOnce` asserts exactly one poll. This also subsumes the earlier timestamp-regression concern, since `timeNow()` is now called under the lock.

**3. Invalid UTF-8 (part of the C1 comment).** The C1 half landed in ab14bd4; the invalid-UTF-8 half was still real. The old implementation short-circuited and returned the input unchanged when nothing was stripped -- preserving raw invalid bytes exactly. `sanitizeTerminal` now always rebuilds, emitting U+FFFD per bad byte, and `validateHandle` rejects invalid UTF-8 up front.

**4. ST1018 lint failure.** Real CI failure, thanks for catching it -- a raw U+009B byte had landed in the fixture. It is now written as the Go escape `\u009b`. (Go has no `\^[` escape, so I used the standard Unicode form.) I also scanned every file under `internal/` and `cmd/` for stray raw control characters: none remain.

**5. Doc log-level claim.** Fixed -- the troubleshooting section now states the two levels separately: registered-key rejections log at info with a reason, unregistered keys at debug with fingerprint only, so a wrong key leaves no record at normal log levels.

Full suite green, `-race` clean on all four touched packages, gofmt/vet clean.

PR #181, opened 2026-09-02
fix(qwkapi): log every request the API turns away

robbiew:
## Why

A sysop auditing the QWK API port (`8666`) could only see successful operations and a couple of failure cases. Everything the API rejected — throttled requests, bad tokens, probe traffic, TLS handshake failures — left no trace, so brute-force attempts and scanning were invisible in `data/logs/vision3.log`.

## What changed

Every request that reaches the port now leaves a record. All messages start with `qwk api`, so `grep '"qwk api'` gives the full picture.

| Level | Event |
|---|---|
| WARN | `qwk api login` `outcome=rateLimited` — login throttle trip, per IP (the brute-force signal) |
| INFO | `qwk api login` `outcome=badRequest` — malformed JSON body |
| INFO | `qwk api auth` `outcome=noToken` / `badToken` — missing, forged, or expired bearer |
| INFO | `qwk api packet` / `qwk api reply` `outcome=rateLimited` — per handle |
| INFO | `qwk api reply` `outcome=readError` |
| WARN | `qwk api reply` `outcome=tooLarge` — over the 16 MiB cap |
| WARN | `qwk api server` — connection-level errors (TLS handshake failures) |
| DEBUG | `qwk api rejected` — `noClientHeader` / `browser` / `htmlAccept` / `unknownPath` / `badMethod` |

Supporting changes:

- **`http.Server.ErrorLog` now routes through slog**, so net/http's own connection errors land as structured JSON records instead of going out through the unstructured stdlib log bridge.
- **Added a `/` catch-all to the mux** — the bare `ServeMux` 404 was invisible, so unknown-path scanning left no trace at all.
- **Guarded `(nil, nil)` service returns** in `handlePacket` / `handleReply`. net/http would recover the panic, but the client saw a dropped connection and the log got a stack trace instead of a usable record; now it's a logged ERROR and a clean 500.

## Judgment calls

**Probe traffic sits at DEBUG.** A public port collects constant `/wp-login.php`-style scanning; at INFO that would bury real activity. Sysops flip `logging.level` to `debug` when they want to see who's knocking. Everything security-relevant that got past the client filter — failed logins, bad tokens, throttling, oversized uploads — is at INFO/WARN and visible at the default level.

**The malformed-login path deliberately omits the handle.** On a decode failure the field may be partial or actually hold the password. There's a test asserting no `handle` field is logged there.

QWK API sessions still do not appear in Last Callers — that list is for terminal logins, and this PR does not change that.

## Testing

New `internal/qwkapi/logging_test.go` — 8 tests that capture `slog` output and assert message, level, and fields for each case (rate limits, auth rejections, probes, bad JSON, oversized upload, `serverErrorWriter`, nil-result guards). Full `go test ./...` passes; `go vet` clean.

## Docs

`docs/sysop/messages/qwk-api.md` gains a **What gets logged** section: the full level/message table, where the log lives, the DEBUG note for probe traffic, and a line confirming API sessions are not in Last Callers.

https://claude.ai/code/session_013PwzJUyWT4fbPgRjbx7K9h

## Summary by CodeRabbit

* **New Features**
  * Added structured logging for QWK API authentication attempts, rejected probes, invalid requests, rate limits, server errors, and oversized responses.
  * Added debug logging that records why rejected probes were declined.
  * Added logging for unknown paths and HTTPS connection-level errors.

* **Documentation**
  * Documented QWK API log destinations, formats, event categories, severity levels, recorded outcomes, and how to enable debug logging for rejected probes.

### DIFF
--- a/docs/sysop/messages/qwk-api.md
+++ b/docs/sysop/messages/qwk-api.md
@@ -85,8 +85,9 @@ Everything the API does goes to the normal BBS log (`data/logs/vision3.log` by
 default, JSON, one record per line) — there is no separate API log file. QWK API
 sessions do **not** appear in Last Callers; that list is for terminal logins.
 
-Every record's `msg` starts with `qwk api`, so `grep '"qwk api'` on the log gives
-you the full picture of API activity.
+Apart from the `QWK API listening` line at startup, every record's `msg` starts
+with `qwk api`, so `grep -i '"qwk api'` on the log gives you the full picture of
+API activity.
 
 | Level | `msg` | When |
 |-------|-------|------|
@@ -100,7 +101,7 @@ you the full picture of API activity.
 | WARN | `qwk api reply` | `outcome=tooLarge` — an upload over the 16 MiB cap. |
 | ERROR | `qwk api build packet` / `qwk api import rep` | A packet failed to build or import. |
 | WARN | `qwk api server` | Connection-level trouble from Go's HTTP server, e.g. `http: TLS handshake error from ...`. |
-| DEBUG | `qwk api rejected` | A request turned away without doing any work: `reason=noClientHeader`, `browser`, `htmlAccept`, `unknownPath`, or `badMethod`. Each carries `remote`, `method`, `path`, and `agent`. |
+| DEBUG | `qwk api rejected` | A request turned away without doing any work: `reason=noClientHeader`, `browser`, `htmlAccept`, `unknownPath`, `badMethod`, or `pathRedirect`. Each carries `remote`, `method`, `path`, and `agent`. |
 
 Failed logins, bad tokens, oversized uploads, and rate-limit trips are all
 recorded at INFO or WARN, so they are visible with the default log level.
--- a/docs/sysop/messages/qwk-api.md
+++ b/docs/sysop/messages/qwk-api.md
@@ -91,7 +91,7 @@ you the full picture of API activity.
 | Level | `msg` | When |
 |-------|-------|------|
 | INFO | `QWK API listening` | Startup: listen address and TLS cert fingerprint. |
-| INFO | `qwk api login` | `outcome=success` / `outcome=fail` / `outcome=badRequest`, with `handle` and `remote`. |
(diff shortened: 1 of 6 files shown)


robbiew:
All three Copilot review comments addressed in 9c34527:

- **`logBadMethod` missing `agent`** — removed the near-duplicate helper; wrong-method rejections now go through `logProbe`, so every `qwk api rejected` record carries the same `remote`/`method`/`path`/`agent` fields. The probe test asserts that shared shape.
- **`serverErrorWriter` CRLF** — now trims `\r\n`; the test feeds a CRLF line and checks neither terminator survives.
- **Docs `badRequest` row** — corrected: that outcome logs `remote` and `error` but deliberately no `handle`, since on a decode failure the field may be partial or hold the password. Also documented which fields every rejection record carries.

robbiew:
CodeRabbit's two findings addressed in 01a3c20:

**`ServeMux` path redirects (real gap — good catch).** Confirmed with a probe: `/api//qwk/login` and `/scan/../..//wp-admin` get a **307 from the mux before dispatch**, so they never reached a registered handler — not even the `/` catch-all — and left no trace. The mux is now wrapped in a handler that records the status and logs `reason=pathRedirect` when the mux deflected the request itself. Regression test covers a doubled slash and a traversal-style path.

**Docs wording.** The grep hint now excludes the startup record (`QWK API listening` doesn't start with `qwk api`) and suggests `-i`. The `badRequest`/`handle` half of that comment was already fixed in 9c34527, before this review landed.

`go test ./...` and `go vet` clean.

robbiew:
Test-rigor finding addressed in 1e87283.

**Probe test** — was checking key *presence* only, and every request shared one User-Agent, so the `agent` field was never genuinely exercised. Rewritten as a table with a distinct method, path, and User-Agent per case, all matched exactly against the logged record. Also added the `htmlAccept` reason, which had no coverage at all.

**`serverErrorWriter` test** — now compares the whole logged value against `strings.TrimRight(line, "\r\n")` rather than a substring, so a writer that dropped detail after the prefix would fail.

The `ServeMux` redirect comment that re-posted on `server.go:78` was already fixed in 01a3c20 (re-review appears to have run against the earlier diff).

`go test ./...` and `go vet` clean.

PR #186, opened 2026-09-03
feat(menu): show Area/Conf/Total/New/Yours in the area lists

robbiew:
The message-area picker listed Tag, Board Name and Description — none of which told a caller whether an area was worth entering. This swaps those for the area name, its conference, and per-area message counts.

**Before**

```
 #   Tag              Board Name     Description
 1   fsxnet_netmail   fsxNet Netmail                              netmail
 2   fsxnet_fsx_gen   General Chat..                              echomail
```

**After** (captured from a live board)

```
     Area                               Conf           Total    New  Yours
-------------------------------------------------------------------------------
 NEW FelonyNet General Chat             FelonyNet        372     12      5
     BBS Support/Dev                    FelonyNet        325      0      2
 NEW Sysops Only Discussion Area        FelonyNet        294      1      4
     FelonyNet Test                     FelonyNet          0      0      0
```

### What changed

- **Columns** — `Area`, `Conf`, `Total`, `New`, `Yours` in both the `SELECTMSGAREA` lightbar picker and the text listing (`LISTMSGAR` and the no-template fallback). The picker flags areas holding unread mail with `NEW` in the left gutter; the text listing keeps its row numbers there, since its `#`/tag prompt needs them.
- **Counts** — `MessageManager.GetAreaCounts` returns total/unread/personal from a single JAM base open, built once per listing rather than per rendered row. "Yours" counts every message addressed to the caller's handle, matched against the recipient CRCs in the `.jdx` index — 8 bytes per message instead of a full header read (`Base.CountMessagesToUser`).
- **Header** — `MSGAREA.TOP` places the column titles with a new `^COLS` token that the BBS renders from the same width constants the row template uses, so the titles cannot drift out of alignment with the data. Templates without the token are drawn unchanged.
- **Fix** — `MSGAREA.TOP.ANS` now ends with a newline after its rule. Without it the first row of the text listing lost its leading space at column 80 and sat one column left. Pre-existing, surfaced while testing this.

New row tokens: `^CF` (conference, 12), `^TM` (total, 7), `^NM` (new, 6), `^YM` (yours, 6), with `^NA` widened to 34. `^TAG`, `^DE` and `^DS` still substitute, so a customized `MSGAREA.MID` keeps working — it just won't show the counts.

### Upgrading an existing install

Rebuild the binaries and copy two templates:

```bash
cp menus/v3/templates/MSGAREA.MID      /path/to/your/bbs/menus/v3/templates/
cp menus/v3/templates/MSGAREA.TOP.ANS  /path/to/your/bbs/menus/v3/templates/
```

Both must move together — the row widths and the rendered header are matched.

### Testing

- New unit tests: JAM recipient counting (incl. the multi-batch `.jdx` scan path) and a layout test that renders the shipped `MSGAREA.MID` against the rendered header, so editing one without the other fails the build.
- `go test ./...` and `go vet ./...` clean.
- Verified end-to-end against a running instance over telnet: both conferences, conference switching with left/right, the `NEW` flags, and the text listing.

https://claude.ai/code/session_012gzWd4rKAdaGMiR9WHUzb6

## Summary by CodeRabbit

* **New Features**
  * Message-area selection now provides a full-screen lightbar picker with conference navigation and read-access validation.
  * Area listings display total, unread, and personal-message counts, including indicators for areas with new messages.
  * A text-list fallback is available when picker templates are unavailable.
  * Area columns now use clearer headers and consistent formatting for conference, time, message number, and year information.

* **Documentation**
  * Updated message-area documentation with column titles, placeholders, picker behavior, fixed-width formatting, and message-count details.


robbiew:
Thanks — all four review points were valid and are fixed in ff40737 / 9f986a5:

- **Header color bleed** (Copilot, `area_counts.go`) — confirmed: the rendered `^COLS` row left the title color open, so the rule line below it came out magenta instead of the dim grey the hand-written header used to end with. The row now closes with `|08`; verified in the rendered bytes (`…Yours\x1b[1;30m` before the rule line) and covered by a test that fails if the header stops ending in a color code.
- **"single pass" wording** (Copilot, `manager_read.go`) — reworded: `GetAreaCounts` opens the base once for all three counts, it does not compute them in one pass.
- **Sequential token substitution** (CodeRabbit, `area_counts.go`) — real: an area named e.g. `^TM` would have been rewritten by the later replacement. Both callers now share `renderAreaRow`, which fills every token in one `strings.NewReplacer` pass; legacy `^TAG`/`^DE`/`^DS` still work. New test `TestRenderAreaRowDoesNotRescanValues` locks it in.
- **`^DS` doc entry** (CodeRabbit, `message-areas.md`) — correct, `^DS` is `area.AreaType`. Documented as the area type, with `^TAG` and `^DE` listed separately.

`gofmt`, `go vet`, `golangci-lint` (0 issues) and `go test ./...` are clean, and the picker was re-checked against a live board after the refactor.

PR #188, opened 2026-09-03
fix(gate): keep challenge gate art border aligned after token substitution

robbiew:
## Problem

On connect, the bot-check frame broke on the prompt line:

```
 +/--------------------------------------------------/+
 | Press ESC 2 times if you're not a bot. |
 | You have 14 seconds.                               |
 +/--------------------------------------------------/+
```

`BOTCHECK.ASC` reserves 23 columns for `{KEY} {PRESSES} {TIMES}`, but the substituted text (`ESC 2 times`) is 11 wide, pulling the right border 12 columns left.

## Fix

`substituteGateTokens` now works line by line and, when a line's width changes, re-pads it in the gap before a trailing box border (`realignBorder`). A trailing run only counts as a border when it is pure frame characters — no letters, digits, or sentence punctuation — so plain sentence lines such as the built-in fallback prompt are left exactly as substituted. If the configured key and press count render wider than the tokens, the gap is absorbed instead.

Rendered from the real art file:

```
 +/--------------------------------------------------/+
 | Press ESC 2 times if you're not a bot.             |
 | You have 20 seconds.                               |
 +/--------------------------------------------------/+
```

Verified the same for `*` / 1 press and for an over-long key (border holds, gap closes up).

## Tests

Added cases to `TestSubstituteGateTokens` for boxed art, the over-wide key, and the unpadded sentence line. `go test ./internal/menu/` and `go vet ./internal/menu/` pass.

Docs updated in `docs/sysop/configuration/security.md`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Challenge Gate prompt substitutions so framed artwork remains properly aligned when replacement values change line widths.
  - Preserved spacing, trailing spaces, line endings, and colored border alignment.
  - Ensured ordinary sentence lines, including fallback prompts, remain unchanged.
  - Improved handling of prompts containing terminal color formatting.

- **Documentation**
  - Updated security documentation to explain alignment behavior for framed artwork and unframed text after prompt substitution.

### DIFF
--- a/docs/sysop/configuration/security.md
+++ b/docs/sysop/configuration/security.md
@@ -109,6 +109,13 @@ correctly-pluralized noun ("time" for a count of 1, "times" otherwise), so
 custom art always reflects the current configuration and renders grammatically
 correct prompts instead of hardcoding values that can drift out of sync.
 
+Because the substituted text is rarely the same width as the tokens it
+replaces, a line that ends in a box border (any trailing run of frame
+characters, such as `|`, separated from the text by a space) is re-padded in
+the gap before that border, so framed art stays square whatever the configured
+key and press count render as. Lines that simply end in words — the built-in
+fallback prompt, for example — are left exactly as substituted.
+
 ### Connection-Rate Limiter
 
 The connection-rate limiter closes a gap the existing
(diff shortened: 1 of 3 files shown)


robbiew:
Addressed the CodeRabbit finding in a8d63fc: `realignBorder` now skips ANSI escape sequences when locating and classifying the trailing border, so art that closes the line with a reset (`... |\x1b[0m`) or colors the border itself (`... \x1b[36m|`) is realigned too, with the escapes re-emitted unchanged after the padding. Regression tests added for both shapes.

PR #189, opened 2026-09-03
feat(editor): split-pane quote picker for CTRL-Q

robbiew:
CTRL-Q printed the numbered source list into the editing area with `\r\n`, which scrolled the header off the top and the footer off the bottom, then asked for a start and an end line with no preview of the result.

Quote mode now splits the editing area, Mystic-style: the reply stays visible above a divider carrying the key legend, and the message being replied to is listed below it under a lightbar.

```
HHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHHH  header template — untouched
Yeah, I ran into the same wall.
--- Bucko Said ---
Bu> Sh> I'd be interested in this as well... Plus, can rlogin send something
Bu> Sh> similar to SBBS's xtrn=? to mystic? I'd love to be able to run mystic
Bu> Sh> doors seamlessly like I can sbbs...
--- Bucko Done ---

─ Quoting Bucko ──────── Up/Dn Pick  SPACE Add  BKSP Undo  TAB Pane  ESC Done ─
  1 On 28 Aug 2026, Shurato said the following...
  2  Sh>
  3  Sh> I'd be interested in this as well... Plus, can rlogin send something
  4  Sh> similar to SBBS's xtrn=? to mystic?  I'd love to be able to run mystic
  5  Sh> doors seamlessly like I can sbbs...
  6  Sh>
  7
  8 I haven't had much luck with it. I got it working, but the script as far as
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF  footer template — untouched
```

Lines 3–5 above were added with three SPACE presses; the top pane grew as they went in. The tearline and `* Origin:` are filtered out of the list.

## Interaction

| Key | Action |
| --- | --- |
| `↑` `↓`, PgUp/PgDn, Home/End | move the bar over the source |
| `SPACE` / `ENTER` | insert the highlighted line and step down |
| `BACKSPACE` | un-quote the last line (drops the banners when the block empties) |
| `TAB` | scroll the reply pane |
| `ESC` | leave quote mode |

Inserting immediately means the reply pane always shows exactly what will be saved, and skipping lines gives non-contiguous quoting — something the range prompt could not do at all. Already-quoted lines are dimmed.

Everything paints inside the rows the editor already owns, one column short of the terminal edge to avoid autowrap on real clients. On exit it does `ClearCache` + `DisplayFooter` rather than `FullRedraw`, so there is no screen-clear flash.

## Bugs fixed along the way

- `SetLine` on the cursor line **destroyed whatever the user had already typed there** — now `InsertLine`.
- `lineText[:75]` / `quotedLine[:79]` byte-sliced, splitting multibyte runes and dropping the tail of long lines — now word-wrapped with the prefix re-applied to each continuation line.
- 99-line cap on the quotable list, removed.
- FTN trailer (tagline, tearline, `* Origin:`, SEEN-BY/PATH kludges) filtered out of the quotable list.
- The source line's own indent is collapsed, so nesting reads `Bu> Sh> …` rather than gaining a space per reply round.
- `quoteTop` / `quoteBottom` / `QuotePrefix` in `templates/configs/strings.json` were declared but never read; they are wired up now. The prefix template gained `^I` (author initials) and defaults to `"^I> "`; a template with no token is used verbatim, so a plain `> ` is still available.

## Testing

14 tests in `internal/editor/quote_test.go`, including a regression test that paints sentinels in the header and footer rows and asserts they survive quote mode. `go build ./...`, `go vet ./...` and `go test ./...` are clean repo-wide.

## Summary by CodeRabbit

- **New Features**
  - Added a split-pane quote picker for selecting and inserting message text.
  - Added keyboard navigation, paging, pane switching, undo, cancellation, and cursor restoration.
  - Added configurable quote banners and prefixes, including author initials and names.
  - Added wrapping, formatting cleanup, and filtering for quoted content.
- **Bug Fixes**
  - Preserved cursor position when quote data is unavailable.
  - Improved prompt and footer handling after quote mode.
- **Documentation**
  - Updated metadata and help text with substitution codes and quote-mode controls.
- **Style**
  - Updated default quote formatting with new delimiters and line prefixes.


robbiew:
Both outside-diff findings from the latest CodeRabbit pass reviewed:

**`quote.go:72` — restoring filtered trailer lines: fixed** in 42eb035. A body that filters down to nothing now shows a "Nothing to quote in that message" notice and stays out of quote mode, rather than falling back to offering the tearline and origin. That also avoids opening a lightbar over an empty list. Covered by `TestQuoteModeWithOnlyTrailerShowsNotice`.

**`runeLen` vs terminal display cells — not doing this here.** The concern is real for CJK, but the fix does not belong in this PR:

- The rest of the editor measures width in *bytes*: `wordwrap.go` compares `len(line)` against `MaxLineLength`, and `buffer.go` does the same. `runeLen` is already strictly more correct than the code it sits next to.
- Changing only the quoter to display cells would make the quoter and the editor disagree about where line 79 ends, which is worse than both being wrong the same way. Doing it properly means converting `wordwrap.go` and `buffer.go` too, plus an East Asian Width table or a new dependency.
- This is a CP437/ANSI BBS talking to an FTN message base. CP437 has no CJK code points at all, and UTF-8 sessions are transcoded through the same tables.

So it is a pre-existing, package-wide property rather than something this PR introduces or regresses — and CodeRabbit labels it a heavy lift itself. Worth its own issue if wide-character support is ever on the roadmap; out of scope for a quote-picker change.

PR #205, opened 2026-09-04
fix(news): show System News at login by default

### LINKED ISSUE #198: Wire up "system news" by default to login sequence
Sysops can add System News items, but they don't display automatically.

robbiew:
Closes #198.

Sysops could add System News items but they never appeared automatically. Two separate things were missing.

## Wiring

`PRINTNEWS` was already a supported login sequence command, and `LISTNEWS`/`EDITNEWS` were already bound on the MAIN and ADMIN menus — but `PRINTNEWS` was absent from both the shipped `templates/configs/login.json` and the built-in default sequence used when `login.json` is missing. It now runs after `NMAILSCAN`.

`NEWSHDR.ANS` opens with `|CL`, so the item clears the screen itself and must **not** set `clear_screen` — doing so would blank the previous step's output even on logins with no news to show. The docs previously claimed `PRINTNEWS` ignores `clear_screen`/`pause_after`; it does not, so that was corrected too.

## Visibility

Even once wired up, non-`always` items could never display. `Authenticate()` stamps `LastLogin` with `time.Now()` before the login sequence runs, so `item.When.After(lastLogin)` was always false by the time `PRINTNEWS` executed.

The prior stamp is now kept as `User.PreviousLogin`, and the three "new since your last visit" checks use it: news, rumors newscan, and the file newscan.

## Seen tracking

For news the date filter is replaced outright with a per-user seen list (`SeenNewsIDs`). A date comparison also silently lost items whenever a caller dropped carrier or fast-logged past the news; tracking IDs means an item that was never displayed is still waiting on the next call.

- `always: false` items show exactly once per user.
- Reading an item from `RUN:LISTNEWS` marks it seen, and drives the `[NEW]` tag.
- `always: true` items are never recorded and still show every login.
- IDs of deleted items are pruned, so the list cannot grow without bound.
- Existing users are back-filled once from `PreviousLogin`, so upgrading systems do not dump the whole backlog.

Seen tracking needs stable IDs, but `newsAddItem` assigned `len(items)+1`, which collides after a delete (add 3, delete 1, add → duplicate ID). It now assigns `max+1`, and `EDITNEWS` repairs existing data on entry. Items with no usable ID fall back to the old date filter.

Access level and max-level filtering are unchanged.

## Note for existing installs

`configs/login.json` is not overwritten on upgrade, so existing systems still need `{"command": "PRINTNEWS"}` added by hand. This is called out in both docs pages.

## Testing

`go build ./...` and `go test ./...` pass. New unit tests in `internal/menu/news_seen_test.go` cover ID allocation after deletion, ID normalization idempotency, seen-set pruning, change detection on store, and back-fill behavior for both upgrading and first-time callers.

Not manually exercised against a live BBS session.

## Summary by CodeRabbit

- **New Features**
  - Login now displays news before other standard login information.
  - News items are tracked individually, allowing unread items to persist across sessions.
  - `Always` news remains visible, while `LISTNEWS` marks items as seen.
  - Login placeholders now support additional user details, activity counts, and prior-login information.
  - New users and migrated accounts receive appropriate news history.

- **Bug Fixes**
  - New-item detection for news, rumors, and file listings now consistently uses the previous login.

- **Documentation**
  - Updated administrator guidance for news tracking, login configuration, migration behavior, and available placeholders.

### DIFF
--- a/cmd/vision3/main.go
+++ b/cmd/vision3/main.go
@@ -1668,6 +1668,7 @@ func main() {
 	if err != nil {
 		logging.Fatal("failed to load login sequence configuration", "error", err)
 	}
+	menu.WarnIfNewsUnwired(rootConfigPath, loginSequence)
 
 	// Initialize session registry for who's online tracking
 	sessionRegistry = session.NewSessionRegistry()
--- a/docs/sysop/menus/menu-system.md
+++ b/docs/sysop/menus/menu-system.md
@@ -323,7 +323,7 @@ Dynamic content placeholders in prompts and ANSI files:
 - `|UPLDS` - User's upload count
 - `|DNLDS` - User's download count
 - `|POSTS` - Number of messages the user has posted
-- `|PV` - Number of users pending validation
+- `|PV` - Number of users pending validation. Only populated on the **MAIN** menu and only for users at CoSysOp level or above; it renders as `0` in every other prompt.
 - `|NODE` - Current node number
 - `|MN` - Current menu name
 - `|GL` - Group/Location (from user profile)
--- a/docs/sysop/menus/menu-system.md
+++ b/docs/sysop/menus/menu-system.md
@@ -709,7 +709,7 @@ Functions available via `RUN:` command:
 
 ### News
 
-- `PRINTNEWS` - Display news items new since last login (login sequence use)
+- `PRINTNEWS` - Display news items the user has not seen yet (login sequence use)
 - `LISTNEWS` - Browse all visible news items; user selects to read
 - `EDITNEWS` - SysOp news management: add, delete, edit, list, view (SysOp only)
 
--- a/docs/sysop/menus/menu-system.md
+++ b/docs/sysop/menus/menu-system.md
@@ -303,6 +303,10 @@ ANSI files contain the visual display for menus. They support:
 Dynamic content placeholders in prompts and ANSI files:
 
 - `|UH` - User handle
+- `|ALIAS` - User handle (alias for `|UH`)
+- `|HANDLE` - User handle (alias for `|UH`)
+- `|NAME` - User's real name
+- `|LEVEL` - User's access level
 - `|TL` - Time left (in minutes)
 - `|CA` - Current message area tag
 - `|CAN` - Current message area name (resolved display name)
@@ -315,6 +319,11 @@ Dynamic content placeholders in prompts and ANSI files:
 - `|DATE` - Current date (MM/DD/YY)
 - `|TIME` - Current time (HH:MM)
 - `|CALLS` - User's total calls
+- `|LCALL` - Date of the user's previous call (MM/DD/YY), or `Never` on a first call
+- `|UPLDS` - User's upload count
+- `|DNLDS` - User's download count
+- `|POSTS` - Number of messages the user has posted
+- `|PV` - Number of users pending validation
 - `|NODE` - Current node number
 - `|MN` - Current menu name
 - `|GL` - Group/Location (from user profile)
--- a/docs/sysop/menus/news.md
+++ b/docs/sysop/menus/news.md
@@ -27,11 +27,21 @@ News items are stored in `data/news.json`. Each item has:
 When `PRINTNEWS` is in the login sequence, ViSiON/3 checks each news item against:
 
 1. **Access level** — user's level must be `>= Level` and `<= MaxLevel` (if set)
-2. **Date filter** — item's `When` must be after the user's `lastLogin`, **or** `Always` is `true`
+2. **Already seen** — the item must not already be in the user's seen list, **or** `Always` is `true`
 
 Each qualifying item is displayed using `NEWSHDR.ANS` (header) followed by the body text. The user presses a key after each item to continue. If no qualifying items exist, the login step is silent.
 
-This matches ViSiON/2's `PrintNews(0, True)` call.
+This matches ViSiON/2's `PrintNews(0, True)` call, with one deliberate difference: V2 decided what was new by comparing the item's date to the user's last login. ViSiON/3 instead records which item **IDs** each user has actually been shown, in `seen_news_ids` on the user record.
+
+The date comparison could not work reliably here — the login timestamp is written at authentication, before the login sequence runs — and it also lost items whenever a caller dropped carrier or fast-logged past the news. With ID tracking, an item that was never displayed is still waiting on the next call.
+
+Consequences:
+
+- An item marked `always: false` is shown **exactly once per user**, whenever that user next completes the news step.
+- Reading an item from `RUN:LISTNEWS` also marks it seen, so it will not reappear at the next login.
+- `always: true` items are never recorded and display on every login.
+- Deleting a news item prunes its ID from every user's seen list on their next news step, so the list cannot grow without bound.
+- On a system upgrading to seen-tracking, each user's list is back-filled once from their previous login timestamp, so existing callers are not shown the entire news backlog.
 
 ---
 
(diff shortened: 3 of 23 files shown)


robbiew:
Added a second commit (757f4a8) fixing `|LCALL`, which had the same root cause as the news date filter.

`|LCALL` is meant to show the caller's *previous* visit, but it read `currentUser.LastLogin` — stamped with `time.Now()` by `Authenticate()` before the session starts — so it always rendered as today's date. It now uses `PreviousLogin`, and a first-time caller correctly keeps the `Never` default.

While in there, two adjacent placeholder bugs:

- `|DNLDS` and `|POSTS` were seeded with a `"0"` default and then never populated from the user record, so every caller saw `0` regardless of their real download and post counts. Now filled from `NumDownloads` and `MessagesPosted`.
- Nine placeholders were implemented but absent from the menu system reference (`|ALIAS`, `|HANDLE`, `|NAME`, `|LEVEL`, `|LCALL`, `|UPLDS`, `|DNLDS`, `|POSTS`, `|PV`). Documented.

The user-derived placeholders are extracted into `applyUserPlaceholders` so this is testable without standing up a terminal — the conference, time-left, and pending-validation placeholders keep their existing inline logic. Three new tests in `internal/menu/display_placeholders_test.go`.

robbiew:
Third commit (018febc) closes the upgrade gap for existing installs.

`PRINTNEWS` now ships in the default login sequence, but all three setup paths (`setup.sh:132`, `dev-setup.sh:141`, `docker-entrypoint.sh:45`) copy a template only `if [ ! -f "$target" ]` — so a system installed before this change keeps its old `configs/login.json` and still silently never displays news.

There's no config migration framework, and auto-appending the step would rewrite a sysop-owned file without being able to distinguish "never had it" from "deliberately removed it". So it reports rather than repairs: if `data/news.json` holds items and the login sequence has no `PRINTNEWS` step, startup logs

```
level=WARN msg="system news items exist but will never be displayed at login"
  items=2 reason="the login sequence has no PRINTNEWS step"
  fix="add {\"command\": \"PRINTNEWS\"} to configs/login.json"
```

Silent when there's no news, and silent when the step is present. Three tests cover fires-when-stranded, silent-when-wired (including a lower-case command), and silent-when-no-news.

Also filed #208 for the one remaining instance of this bug class that I deliberately left out of this PR: the scripting bridge exposes `lastLogin` with no `previousLogin` counterpart, so scripts asking "what's new since this user was last here?" hit the same trap. That's an API surface addition rather than a fix.

robbiew:
## Verified end-to-end against a live BBS

Ran this against a real install (`~/bbs-dev`, telnet 2323) driving actual login sessions over a socket — not just unit tests. Then rebuilt the PR base commit (`f726d6e`) and ran it against the **same install with identical seeded state** to confirm the counterfactual.

### Before vs after, same install, same starting state

Seeded: `lastLogin` backdated to 2026-07-04, 42 downloads, 17 posts, four news items exercising each filter.

| | OLD (`f726d6e`) | NEW (this PR) |
|---|---|---|
| `\|LCALL` | `09/04/26` — today | `07/04/26` — actual last call |
| `\|DNLDS` | `0` | `42` |
| `\|POSTS` | `0` | `17` |
| Non-`always` news items shown | **none** | correct set |

The old binary suppressed every non-`always` news item, exactly as predicted — `PRINTNEWS` was effectively dead code for anything but `always: true`.

### Filter matrix (new binary, first login)

| Item | Condition | Expected | Result |
|---|---|---|---|
| Welcome To The Board | dated after previous login | show | ✅ shown |
| Old Announcement | dated before previous login | hide (back-filled) | ✅ hidden |
| Always Read This | `always: true` | show every login | ✅ shown |
| New Users Only | `max_level: 100`, user is 255 | hide | ✅ hidden |

`NEWSHDR.ANS` rendered with every substitution resolved — `^NM`/`^TI`/`^FR`/`^DT`/`^TM`/`^LV`/`^MX` — including `Level: 0-All` for the unset max. No empty-parenthesis artifacts.

### Persistence and seen tracking

After first login:
```
lastLogin     = 2026-09-04T09:24:49   (this session)
previousLogin = 2026-09-03T10:24:43   (the real prior visit)
seen_news_ids = [1, 2]                (2 back-filled, 1 marked after display)
```
Second login showed **only** the `always` item — item 1 correctly suppressed.

### The abort case

This is the scenario the old date-based logic loses, so I tested it directly:

1. Added a new item dated 09:00.
2. Authenticated, then **dropped carrier before the news step**. `previousLogin` advanced to 09:26 — now *past* the item's timestamp.
3. Reconnected.

The item still displayed, then recorded as seen (`[1, 2, 5]`). Under the old logic it would have been lost permanently.

### Startup warning

Fires against the real install with news present and no `PRINTNEWS` step, and clears once wired:
```
level=WARN msg="system news items exist but will never be displayed at login"
  items=4 reason="the login sequence has no PRINTNEWS step"
  fix="add {\"command\": \"PRINTNEWS\"} to configs/login.json"
```

### Docs (commit 8d860eb)

Verification surfaced sysop docs still describing the old behavior. `user-management.md` was the biggest gap — the user record reference documented `lastLogin` with no hint it is stamped at authentication. Added the new fields plus a section on which subsystems measure against `previousLogin`. Also corrected `rumors.md`, both command reference tables, added a news troubleshooting section, and flagged the trap in `vpl-scripting.md` for #208.

The test install was restored to its original state afterward (user record, menu prompt, test news removed); `PRINTNEWS` was left in its `login.json` since that is the intended end state.

robbiew:
Addressed the review in `b6468e9`. Both Major findings were real — verified each against the code before changing anything.

**ID reuse (CodeRabbit, Major) — confirmed.** Reproduced it: with items `[3,2,1]`, deleting ID 3 made `nextNewsID` return 3 again. A user who had seen the deleted item still carries that ID until their next news step, so the replacement was treated as seen and hidden from them permanently. `NewsData` now persists a monotonic `NextID`, and allocation takes the live items only as a *floor* so a `news.json` written before the field existed cannot collide on first use. `normalizeNewsIDs` draws from the same allocator rather than filling the lowest free gap — a gap usually means a deleted item, so filling it recreates the same bug.

**Authenticate snapshot race (CodeRabbit, Major) — confirmed.** Fixed by copying `*user` under the lock that wrote the fields, instead of re-reading after unlock. Added a concurrent test that asserts each session's snapshot is internally consistent (`PreviousLogin` strictly before its own `LastLogin`); passes under `-race`.

**Normalize on load in PRINTNEWS/LISTNEWS (Copilot) — valid, done.** I had handled `ID <= 0` but not duplicates, which would collapse several items under one seen-key. Both read paths now normalize in memory. Normalization is deterministic for a given file, so sessions agree with each other and with what the sysop editor eventually persists — no user session writes `news.json`.

**LISTNEWS missing the back-fill (Copilot) — valid, done.** A user reaching the list before ever reaching PRINTNEWS saw the whole backlog tagged `[NEW]`, and the marker was never persisted for them. It now runs the same one-time back-fill and saves it even when nothing is read.

**Regression tests for the newscan cutoff (Copilot) — valid, done.** Rather than test the two call sites separately, I extracted `newscanSince`/`isNewSince` so the PreviousLogin semantic is stated once and used by news, rumors, and the file newscan. Tests assert the cutoff is *not* `LastLogin`, cover the zero/first-caller case, and check the boundary (posted exactly at the cutoff is not new).

**Docs** — all three applied: `|PV` now notes it is populated only on MAIN for CoSysOp and above, the `PRINTNEWS` reference row mentions Always items, and the bare fences are tagged `text` (including one pre-existing MD040 in the same file).

Re-verified live against a real BBS after the allocator rewrite — first login shows the three qualifying items, second login shows only the `always` one, and `seen_news_ids` persists as `[1, 2, 5]`. Full `go build ./...` / `go test ./...` clean, `go vet` clean.

Issue #209, opened 2026-09-04
Templates need a way to drop decoration around an empty placeholder


robbiew:
Split out of #201 as the underlying mechanism, so that issue can stay an audit of existing templates.

## Problem

Templates embed literal decoration — a label, parentheses, a separator — around a placeholder, and there is no way to say "omit the decoration if the value is empty." So an unset field leaves the decoration stranded.

`MAIN.MNU`:

```json
"PROMPT1": "|10|UH |12from |GL (|04|UN|12)|07"
```

renders for any user without a `privateNote`:

```
Felonius from ViSiON/3 ()
```

Nine of the fifteen message header templates have the same shape with `@U@`, producing a labelled row with nothing after it:

```
User Note: 
```

Both fields (`|UN`, `@U@`) are the SysOp-only `privateNote`, which is why this survived testing: a SysOp checking their own account always has it populated, and never sees the empty form.

## Why this is separate from #201

#201 can fix every template shipping today and the problem still returns, because nothing stops the next template being authored the same way. #201 notes these "were not thoroughly tested when they were created" — without a mechanism, that stays true for whatever is created next.

## Proposed change

An optional-group syntax: a delimited span that is dropped entirely when every placeholder inside it resolves to empty, and otherwise renders normally. Strawman using pipe-code style so it cannot collide with existing `|XX` codes:

```
|10|UH |12from |GL |{(|04|UN|12)|}|07
```

Points to settle:

- **Delimiter choice.** Must not collide with existing `|XX` pipe codes, `@X@` AT-codes, or the `@T####…####@` width-padding form. Menu prompts and header templates use different substitution paths (`internal/menu/executor_display.go` vs `internal/menu/message_reader_subst.go`), so the syntax should be picked once and applied to both rather than diverging.
- **"Empty" definition.** Whitespace-only should probably count as empty; a placeholder with a non-empty default like `|CC` → `"None"` never triggers it either way.
- **Nesting.** Probably not needed; flat groups cover every case found in #201.
- **Width padding.** Header templates pad with `@T####…####@`; a dropped group must not leave the surrounding fixed-column layout misaligned. This is the fiddly part — header templates use absolute cursor positioning, so "drop" may need to mean "render as spaces" there while meaning "remove" in a flowing prompt.

## Acceptance

- `MAIN.MNU` renders `Felonius from ViSiON/3` with no stray parens when `privateNote` is empty, and `Felonius from ViSiON/3 (SysOp)` when it is set.
[...]

Issue #201, opened 2026-09-04
Audit Header Templates


robbiew:
I've noticed quite a few selectable "Message Header" templates are malformed, missing data or something display empty things like "()" when there's no data in the replacement code. We need to tighten these up. They were not throughly tested when they were created.

robbiew:
Scoping findings from auditing the templates. Four concrete defects, all verified against the shipped `menus/v3` set.

### 1. `MSGHDR.15` ships but can never be kept

`discoverMessageHeaders` globs `MSGHDR.*.ans` and sorts numerically, so style 15 appears in the selector and can be picked. But `internal/menu/executor_messages_read.go:53` gates on:

```go
if currentUser.MsgHdr < 1 || currentUser.MsgHdr > 14 {
```

Pick style 15 and the reader treats it as unset and re-prompts on **every read**. `internal/user/user.go:77` also documents the range as `1-14`, so either the template is an orphan or the bound is stale. Whichever it is, the two disagree today.

### 2. Two templates never show the subject

Surveying the `@X@` codes actually present in each file (accounting for the `@T####…####@` padding form):

| Template | Missing |
|---|---|
| `MSGHDR.3.ans` | no `@T@` — subject never displayed |
| `MSGHDR.9.ans` | no `@T@` — subject never displayed |

Every template does have `@F@`, so From is fine throughout.

### 3. Nine of fifteen templates print a `@U@` row that is empty for almost everyone

`@U@` is the author's `privateNote` — a **SysOp-only** field. It is set on sysop accounts and empty for essentially every ordinary user, so templates 1, 2, 3, 4, 7, 8, 9, 10, 12 and 15 render a labelled row with nothing after it:

```
User Note: 
```

This is easy to miss precisely because a SysOp testing on their own account always has the field populated. (Related: `message_reader.go:215` already computes `templateUsesUserNote` to vary the From string, so there is partial awareness of this in the code.)

### 4. Same bug outside message headers: `MAIN.MNU`

The menu prompt has the same shape:

```json
"PROMPT1": "|10|UH |12from |GL (|04|UN|12)|07"
```

`|UN` is the same SysOp-only `privateNote`, so every ordinary user sees:

```
Felonius from ViSiON/3 ()
```

[...]

PR #210, opened 2026-09-04
fix(menu): drop prompt decoration when the placeholder is empty

robbiew:
Refs #201, closes the menu-prompt half of #209.

## The bug

`MAIN.MNU` wraps `|UN` in literal parentheses:

```json
"PROMPT1": "|10|UH |12from |GL (|04|UN|12)|07"
```

`|UN` is `privateNote` — a **SysOp-only** field. It is set on sysop accounts and empty for essentially every ordinary user, so the common case renders:

```
Felonius from ViSiON/3 ()
```

A SysOp checking their own account always has the field populated and never sees it, which is how this survived. `|GL` has the same problem one field over, leaving a dangling `from` for a user with no group/location.

I checked all 14 menus with prompts — `MAIN.MNU` is the only affected one. The others wrap `|TIME`, `|TL`, `|CC`, `|CAN`, `|FCN`, `|CFAN`, all of which have non-empty defaults (`"None"`) and cannot produce the artifact. So there is no separate menu audit to do; this is the whole menu-side finding.

## The fix

Editing the template alone would not fix the class — nothing stops the next one being authored the same way, which is exactly what #201 describes ("not thoroughly tested when they were created"). So this adds the mechanism from #209 for prompts.

`|{ ... |}` marks an optional group: dropped in full when every placeholder inside it is empty, rendered normally otherwise.

```json
"PROMPT1": "|10|UH|{|12 from |GL|}|{ (|04|UN|12)|}|07"
```

| User record | Renders |
| --- | --- |
| note and location set | `Felonius from ViSiON/3 (SysOp)` |
| note empty | `Felonius from ViSiON/3` |
| both empty | `Felonius` |

Groups resolve *before* substitution, while the placeholder tokens are still present to test. Design choices, all covered by tests:

- Removed only when the group holds **at least one** placeholder and **all** of them are empty.
- Whitespace counts as empty, so a value padded to a fixed width does not keep a group alive.
- A group containing no placeholder is **kept** — dropping it would silently eat literal text.
- Placeholders with non-empty defaults (`|CC`, `|CAN`, `|TL`, `|LCALL`…) never trigger removal.
- Groups do not nest; an unmatched `|{` degrades to visible text rather than swallowing the rest of the prompt.
- `{` and `}` are not valid pipe-code characters, so the delimiters cannot collide with `|00`-`|15` or `|CL`/`|CR`/`|PP`.

## Scope

This is the **menu-prompt half** of #209. Message header templates use a separate substitution path (`message_reader_subst.go`) with absolute cursor positioning, where "drop" likely needs to mean "render as spaces" to preserve column alignment. That half stays open — nine of the fifteen header templates have the same `@U@` problem.

## Testing

`go build ./...` and `go test ./...` pass. Unit tests cover eleven cases of group expansion plus the expand-then-substitute pipeline.

Verified live against a real BBS over telnet, toggling the two fields on the account and reading the rendered prompt back:

```
note='SysOp'  loc='ViSiON/3'  ->  Felonius from ViSiON/3 (SysOp)
note=''       loc='ViSiON/3'  ->  Felonius from ViSiON/3
note=''       loc=''          ->  Felonius
```

## Note for existing installs

`menus/` is copied at setup and not overwritten on upgrade, so existing systems keep their current `MAIN.MNU`. The `|{...|}` syntax works as soon as the binary is updated; the prompt itself needs the one-line edit. Documented in the menu system reference.

## Summary by CodeRabbit

- **New Features**
  - Added optional prompt groups using `|{ ... |}`. Groups are omitted when contained placeholders are blank and displayed when values are present.
  - Preserved coordinate markers, literal content, unmatched groups, and placeholder substitution behavior.

- **Bug Fixes**
  - Corrected handling of ANSI includes so `|XX` placeholders remain literal text instead of producing errors.

- **Documentation**
  - Documented optional prompt-group syntax, rendering rules, limitations, and ANSI include behavior.

- **Updates**
  - Updated the main menu prompt formatting for clearer source and username display.

### DIFF
--- a/docs/sysop/menus/menu-system.md
+++ b/docs/sysop/menus/menu-system.md
@@ -353,6 +353,13 @@ Rules:
 - Groups do not nest. An unmatched `|{` is left as visible text rather than
   swallowing the rest of the prompt.
 
+Groups are resolved on the prompt string itself, before `%%file.ans%%` includes
+are expanded — so `|{...|}` works in `PROMPT1`/`PROMPT2` but not inside an
+included ANSI file.
+
+The `|{P}` and `|{O}` login position markers share the `|{` prefix but are not
+groups; they are stepped over and pass through untouched.
+
 > Currently available in menu prompts. Message header templates use a separate
 > substitution path and do not support this yet — see issue #209.
 
--- a/docs/sysop/menus/menu-system.md
+++ b/docs/sysop/menus/menu-system.md
@@ -355,7 +355,9 @@ Rules:
 
 Groups are resolved on the prompt string itself, before `%%file.ans%%` includes
 are expanded — so `|{...|}` works in `PROMPT1`/`PROMPT2` but not inside an
-included ANSI file.
+included ANSI file. This is not specific to groups: **no** `|XX` placeholder
+expands inside an included file, and the markup renders as literal text rather
+than failing loudly. See issue #211.
 
 The `|{P}` and `|{O}` login position markers share the `|{` prefix but are not
 groups; they are stepped over and pass through untouched.
--- a/docs/sysop/menus/menu-system.md
+++ b/docs/sysop/menus/menu-system.md
@@ -321,6 +321,41 @@ Dynamic content placeholders in prompts and ANSI files:
 - `|UN` - User note (privateNote from user profile)
 - `|NEWUSERS` - New user registration status (`YES` or `NO`)
 
+#### Optional Groups
+
+Wrap part of a prompt in `|{ ... |}` to drop it when the placeholders inside it
+are empty. Use this for decoration that should disappear along with its value —
+parentheses, a label, a separator — instead of being left stranded.
+
+```json
+"PROMPT1": "|10|UH|{|12 from |GL|}|{ (|04|UN|12)|}|07"
+```
+
+| User record | Renders |
+| ----------- | ------- |
+| location and note both set | `Felonius from ViSiON/3 (SysOp)` |
+| note empty | `Felonius from ViSiON/3` |
+| both empty | `Felonius` |
+
+Without the groups, a user with no note sees `Felonius from ViSiON/3 ()`.
+
+This matters most for `|UN` (the SysOp-only `privateNote`) and `|GL`
+(Group/Location), which are empty for most ordinary users. A SysOp testing on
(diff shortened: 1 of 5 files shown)


robbiew:
Addressed the review in `9ec9529`. Both findings were real, and the first was severe — thanks Copilot, this was my bug.

**`|{P}` / `|{O}` collision — confirmed, and worse than reported.** I claimed in the original commit that `{`/`}` are not valid pipe-code characters. That was wrong: they are login position markers, handled at `internal/menu/conditional_ansi.go:40` and documented in the menu reference. Reproduced:

```
expandOptionalGroups("|{P}user |{(|UN)|}")  ->  ""
```

The marker's `|{` paired with the group's `|}` and deleted the entire prompt. The scanner now steps over markers when looking for an opener, matching the same shape the renderer itself uses. Six regression cases cover marker-then-group, marker-then-kept-group, marker inside a group, marker after a group, and both markers alone.

Worth noting for scope: no file in the shipped menu set uses `|{P}`/`|{O}`, so the default configuration was never affected — but the markers are documented, so a sysop's own `LOGIN.ANS` could have hit it.

**Prefix collision — confirmed.** `groupIsEmpty` tested every placeholder whose name appeared in the group text, so a group containing `|CAN` was also judged on `|CA`. With `|CAN` empty and `|CA` populated, the group stayed visible:

```
expandOptionalGroups("a|{[|CAN]|}b")  ->  "a[|CAN]b"   (want "ab")
```

Now matches longest-key-first and consumes each match, mirroring how substitution resolves the same collision. Tested with `|CA`/`|CAN` and `|CC`/`|CCN` in both directions, plus a mixed group.

**Also documented** something the review prompted me to check: groups resolve *before* `%%file.ans%%` includes are expanded, so `|{...|}` works in `PROMPT1`/`PROMPT2` but not inside an included ANSI file. That is a real limitation and was previously unstated.

Separately, I checked whether any art/ANSI file needs updating for either PR: none do. `NEWSHDR.ANS` uses exactly the seven codes `displayNewsItem` substitutes; `|LCALL`, `|DNLDS`, `|POSTS` and `|UPLDS` appear in no art file at all; and the only other `|GL` user, `USERLIST.MID`, is space-separated with no decoration to strand.

PR #212, opened 2026-09-04
fix(news): word-wrap news bodies instead of letting the terminal split words

robbiew:
`displayNewsItem` wrote each stored body line straight to the terminal with no wrapping:

```go
for _, line := range strings.Split(item.Body, "\n") {
    wv(terminal, strings.TrimRight(line, "\r")+"\r\n", outputMode)
}
```

So anything longer than the screen was hard-wrapped by the terminal at its own margin — mid-word. The message reader has always wrapped properly (`message_reader.go:254`); news simply never did.

## Before / after, same body, same terminal

A 392-character single-line body, captured live over telnet:

**Before** — one line, terminal chops it wherever it lands:
```text
[392] This is a deliberately long paragraph written as one single line with no embedded newlines whatsoever, so that the only thing that can possibly break it across the screen is either our word wrapping or the terminal's own hard margin, and incorporating extraordinarily lengthy vocabulary such as counterrevolutionary and incomprehensibilities to make any mid-word breakage obvious at a glance.
```

**After** — wrapped on word boundaries, every line within budget:
```text
[ 72] This is a deliberately long paragraph written as one single line with no
[ 79] embedded newlines whatsoever, so that the only thing that can possibly break it
[ 72] across the screen is either our word wrapping or the terminal's own hard
[ 68] margin, and incorporating extraordinarily lengthy vocabulary such as
[ 76] counterrevolutionary and incomprehensibilities to make any mid-word breakage
[ 20] obvious at a glance.
```

## Two details the naive version gets wrong

**Pipe codes must be converted before wrapping.** News bodies support colour — `wv` runs them through `ReplacePipeCodes` — so this is a live path, not a theoretical one. `wrapAnsiString` measures width with ANSI escapes stripped but knows nothing about pipe codes, so wrapping the raw body counts `|04` as three visible columns. Demonstrated: a body of **59 visible columns** was being split into **49 + 9**. Converting first (exactly what the reader does at `message_reader.go:234`) fixes it, and the wrapped lines are then written straight through rather than pipe-processed a second time.

**Width is one column short of the terminal.** The body is written as a sequence of CRLF-terminated lines, so a line filling the full width triggers the terminal's own auto-wrap and pushes the CRLF onto the next row — a blank line between every wrapped line. The codebase already reserves this column elsewhere for sequentially written output (`message_reader_subst.go:240`, `v3net_areas.go:173`). The reader can use the full width because it renders into a cursor-addressed region instead.

When the width is unknown this falls back to 79 rather than to "do not wrap" — no wrapping is the original bug.

## Scope

`termWidth` is threaded through `displayNewsItem` to its six call sites, all of which already had it in scope. ANSI art is untouched: `wrapAnsiString` detects cursor positioning and splits on newlines only, so a sysop pasting art into an item still gets it positioned rather than reflowed.

## Testing

`go build ./...`, `go vet`, `go test ./...` all clean. Nine unit tests cover the width budget, CRLF normalization, word-boundary wrapping with exact round-trip of the text, preservation of authored paragraph breaks, ANSI-art passthrough, visible-column measurement with pipe codes, and colour survival.

Verified live against a real BBS for both the plain long paragraph above and a colour-coded one — wraps on visible columns, ANSI escapes intact in the raw stream, no literal pipe codes leaking.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved news display wrapping to fit the available terminal and header width.
  - Split oversized unbroken text, including long URLs, within the visible-width limit.
  - Preserved colors and ANSI formatting while wrapping news content.
  - Kept ANSI artwork layout intact during rendering.
  - Normalized line endings and retained intentional blank lines.
  - Reduced unwanted terminal auto-wrapping artifacts, including for colored and long news content.
  - Improved consistency across news viewing and editing screens.

### DIFF
--- a/internal/menu/news.go
+++ b/internal/menu/news.go
@@ -213,7 +213,7 @@ func WarnIfNewsUnwired(rootConfigPath string, loginSequence []config.LoginItem) 
 //
 //	^NM = item number   ^TI = title    ^FR = from/author
 //	^DT = date          ^TM = time     ^LV = min level   ^MX = max level
-func displayNewsItem(e *MenuExecutor, terminal *term.Terminal, item *NewsItem, idx int, outputMode ansi.OutputMode) {
+func displayNewsItem(e *MenuExecutor, terminal *term.Terminal, item *NewsItem, idx int, outputMode ansi.OutputMode, termWidth int) {
 	ansiPath := filepath.Join(e.MenuSetPath, "ansi", "NEWSHDR.ANS")
 	if raw, err := os.ReadFile(ansiPath); err == nil {
 		maxStr := strconv.Itoa(item.MaxLevel)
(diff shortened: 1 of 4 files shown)


robbiew:
Addressed in `04f3bdf`. The finding was real — reproduced before fixing.

A 139-column URL went out as a single 139-column line for the client terminal to chop mid-token, which is precisely the failure this PR exists to prevent:

```text
line 0:  12 cols
line 1: 139 cols   <-- over the 79 budget, terminal chops it
line 2:   6 cols
```

An over-long token has to break somewhere. Doing it here makes the break point deterministic and keeps it at our budget rather than the terminal's margin — which also avoids the trailing-column auto-wrap the width calculation otherwise guards against. The pass is ANSI-aware: escapes are copied through without counting toward the width and are never split across a chunk boundary.

**One thing the suggestion would have broken, so I gated it.** `wrapAnsiString` deliberately leaves ANSI art unwrapped because art rows are positioned absolutely. Applying the break pass to art would chop full-width rows and push everything below them down a line. I checked rather than assumed — an 80-column art fixture went from 2 rows to 4:

```text
wrapAnsiString left art as 2 lines (correct)
after breakOversizedLines: 4 lines <-- ART CORRUPTED
```

So the break pass runs only when `containsAnsiArt` is false, with a test pinning that gate.

Verified live on a real BBS — the same 139-column URL now arrives as 79 + 60 columns, nothing over budget, and reassembles intact:

```text
[ 21] Grab the archive from
[ 79] https://example.com/downloads/abcdefghijabcdefghijabcdefghijabcdefghijabcdefghi
[ 60] jabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij/file.zip
[ 38] and let me know if the mirror is slow.

lines over the 79-col budget: 0
```

Four new tests as requested: an unbroken token longer than 79 visible columns (asserting every line fits and the URL survives reassembly), short lines untouched, ANSI-awareness (no split escapes, no lost columns), and invalid widths as a no-op.

Note this same gap exists in the message reader, which shares `wrapAnsiString` — a long URL in a message body overflows there too. I left that alone rather than change shared wrapping behaviour in a news PR; happy to file it separately if you'd like.

robbiew:
Verified this one rather than applying it, and I don't think the per-line exemption is safe. The finding is accurate, but the proposed fix trades a rare overflow for a worse failure.

Built the exact mixed body described — positioned banner, a full-width art row, and prose containing a 138-column URL:

```text
containsAnsiArt(body) = true
wrapAnsiString -> 3 lines (art: split on newlines only)
  line 0:   6 visible cols, has-positioning=true
  line 1:  80 visible cols, has-positioning=false   <-- art row
  line 2: 138 visible cols, has-positioning=false   <-- prose + URL
per-line would break the 80-col block row into 2 pieces
```

The art row on line 1 relies on newline flow and carries **no escape of its own**, so a per-line check cannot distinguish it from the prose on line 2. Applying the exemption per line fixes the URL and simultaneously splits that art row in half, pushing everything below it down a line — the exact corruption the body-wide gate exists to prevent.

I looked for a better discriminator and could not find a reliable one. "Has a positioning escape" fails as shown. "Contains whitespace" fails too, since most ANSI art uses spaces for background fill, so it would break art rows as readily as prose.

Current behaviour is the conservative one: a body containing art is left entirely alone, so art is never corrupted, and a long URL in such a body may still be split by the client terminal. That is a cosmetic problem in a rare body shape; the alternative is a structural problem in a common one.

Worth noting the same body-wide decision already exists upstream in `wrapAnsiString`, which does not word-wrap *any* line of an art-containing body — shared with the message reader. Making mixed bodies fully correct means teaching that function per-line semantics, which changes message rendering too and belongs in its own PR rather than this one.

Happy to implement per-line breaking if you'd still prefer it, but flagging the tradeoff rather than applying it silently.

PR #214, opened 2026-09-05
fix(user): stop the BBS reverting sysop edits made in ./ue

### LINKED ISSUE #199: Investigate hot-reloading of changes to user data when using ./ue TUI
When making a change to a user (e.g. change level, validate) using the ./ue TUI, it doesn't appear those changes are loaded.

robbiew:
Closes #199.

## It is data loss, not a stale read

The issue reports that edits made in `./ue` "don't appear to be loaded." They are worse than that — they are destroyed. Reproduced end to end against a running BBS:

```text
after ue edit:   level=100  loc=EDITED-BY-UE  validated=false
after a login:   level=255  loc=ViSiON/3      validated=true    *** OVERWRITTEN ***
```

The BBS holds every user in memory for the life of the process and rewrites the whole of `users.json` on each save, and `users.json` is not among the files the config watcher reloads. So any save writes back a copy read at startup.

`./ue` already guards the other direction — it records the file's mtime at load and calls `CheckFileChanged` before saving (`usereditor/model.go:1226`), so it will not clobber the BBS. The BBS had no equivalent. This gives it the same discipline.

## The fix, in two parts

**1. Detect the external edit.** Track the file's mtime across load and save. When it no longer matches, fold the on-disk record in before writing: sysop-owned fields come from disk, the session keeps its own state. Users added externally are adopted; users deleted externally stay deleted.

The sysop-owned set is deliberately the editable field list from `internal/usereditor/fields.go` rather than one invented here, so the editor and the merge cannot disagree about who owns what. Display-only entries there (last login, created/updated stamps, deletion state) are the BBS's to maintain and stay session-owned.

**2. A generation counter, because the mtime check alone was not enough.** The unit tests passed and the live login still failed. Instrumenting the save path showed why:

```text
DEBUG external edit detected, merging
DEBUG about to write  level=100  loc=EDITED-BY-UE     <- merge worked
DEBUG no external change; writing memory as-is
DEBUG about to write  level=255  loc=ViSiON/3         <- reverted 0.2ms later
```

The login sequence calls `UpdateUser` with a copy the session took *before* the merge, and because the manager had just written the file itself, no external change is detected the second time. All 54 `UpdateUser` call sites have this shape: take a copy at session start, mutate it, write the lot back.

So records carry an unexported generation counter (never serialized), bumped when an external edit is folded in. `UpdateUser` and `UpdateUserByID` compare the caller's copy against the authoritative record and, if the copy predates a refresh, restore the sysop-owned fields while keeping the caller's session state.

Sessions can still change those fields normally — new user validation promoting `AccessLevel` and the user config menu changing terminal preferences both depend on that, and both are covered by a test.

## Known cost of the ownership rule

An external edit landing mid-session also takes `TimesCalled` and `FilePoints` from disk, discarding that session's increment. Losing a login count is trivial next to reverting a validation, and one rule with no exceptions is easier to keep correct than a list of them. Flagging it rather than burying it.

## On guarding against editing an online user

Considered and deliberately not added:

- `./ue` cannot currently tell. Online state lives in `activeUserIDs` inside the running BBS process (`manager_stats.go:39`) and is never persisted, so a guard would first need the BBS to publish node state to disk.
- Blocking would break the main use case — validating a new caller *during* their first login is exactly when a sysop reaches for the editor.
- With this merge policy the write is already safe, so a guard is not load-bearing.

There is a real remaining gap, but it is a different one: the running session holds its own copy, so a demotion persists to disk yet does not affect the *current* session until next login. Filed separately rather than conflated with persistence.

## Testing

`go build ./...`, `go vet`, and `go test ./... -race` all clean. Nine tests cover the external edit surviving a save, session state surviving the merge, externally added users being adopted, externally deleted users staying deleted, no-op behaviour when nothing changed externally, the manager not mistaking its own write for an external edit, a corrupt file not costing the session its state, the stale-copy revert that the first version missed, and sessions still being able to change sysop fields normally.

Verified live on a real BBS: the demotion now survives the login and `lastLogin` is still recorded.

## Summary by CodeRabbit

* **Bug Fixes**
  * External edits to the user file are now detected reliably and merged before saves.
  * User additions, updates, and deletions made outside the application are preserved correctly.
  * Session-owned activity data is protected from stale user records.
  * Newly adopted users receive valid identifiers, and legacy usernames are retained as handles.
  * In-memory registrations are preserved during concurrent external edits.
  * Corrupt or unreadable user files no longer replace current data.
  * Unnecessary reloads are avoided when the file is unchanged.

### DIFF
--- a/internal/user/external_edit.go
+++ b/internal/user/external_edit.go
@@ -1,0 +1,144 @@
+package user
+
+import (
+	"encoding/json"
+	"os"
+	"strings"
+	"time"
+)
+
+// The BBS keeps every user in memory for the life of the process and rewrites
+// the whole of users.json on each save. ./ue edits the same file directly from
+// a separate process. Without the merge in this file, a sysop who changes a
+// user's level or validates them while the BBS is running has that edit
+// silently reverted by the next login, because the BBS writes back a copy it
+// read at startup.
+//
+// ./ue already guards the other direction: it records the file's mtime at load
+// and calls CheckFileChanged before saving, so it will not clobber the BBS.
+// This gives the BBS the same discipline.
+
+// sysopOwnedFields copies the fields that ./ue exposes for editing from src
+// onto dst. These are the sysop's to set, so when the file has been edited
+// underneath a running BBS, the on-disk values win over whatever the session
+// happened to be carrying.
+//
+// The set is deliberately the editable field list from
+// internal/usereditor/fields.go — display-only entries there (last login, call
+// counts, created/updated stamps, deletion state) are the BBS's to maintain
+// and are therefore left alone, taken from the in-memory record instead.
+//
+// Anything not listed here stays session-owned. When adding a field to the
+// user editor, add it here too, or an external edit to it will be lost the
+// same way this fixes.
+func sysopOwnedFields(dst, src *User) {
+	// Identity and access
+	dst.RealName = src.RealName
+	dst.AccessLevel = src.AccessLevel
+	dst.Flags = src.Flags
+	dst.Validated = src.Validated
+	dst.TimeLimit = src.TimeLimit
+	dst.FilePoints = src.FilePoints
+	dst.TimesCalled = src.TimesCalled
+	dst.GroupLocation = src.GroupLocation
+	dst.PrivateNote = src.PrivateNote
+
+	// Credentials
+	dst.PasswordHash = src.PasswordHash
+	dst.PublicKeys = src.PublicKeys
+
+	// Terminal and display preferences
+	dst.ScreenWidth = src.ScreenWidth
+	dst.ScreenHeight = src.ScreenHeight
+	dst.PreferredEncoding = src.PreferredEncoding
+	dst.OutputMode = src.OutputMode
+	dst.MsgHdr = src.MsgHdr
+	dst.HotKeys = src.HotKeys
+	dst.MorePrompts = src.MorePrompts
+	dst.CustomPrompt = src.CustomPrompt
+}
+
+// readUsersFromDisk parses users.json into a handle-keyed map without touching
+// the manager's own state. Used to see what an external editor wrote.
+func readUsersFromDisk(path string) (map[string]*User, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	var list []*User
+	if err := json.Unmarshal(StripUTF8BOM(data), &list); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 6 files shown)


robbiew:
All seven findings addressed in `fd1ce71`, each verified against the code first. Six fixed here, one filed separately.

**`TimesCalled` session-owned (Major) — agreed, changed.** I had documented this as an accepted cost, but the reviewer is right that it is the wrong trade: the BBS advances these every login and every transfer, so *any* external edit discarded that session's activity, however unrelated the edited field. A sysop setting them by hand mid-session is far rarer and visible to them when it does not stick. `FilePoints` moved with it for the same reason. My comment claiming call counts were display-only in the editor was also just wrong — `Total Calls` is editable — and is corrected.

**mtime is not change identity (Major) — agreed, replaced.** Now fingerprints contents: size plus a SHA-256 of the bytes. `users.json` is small, so the extra read costs nothing next to the write that follows. New test pins the mtime with `os.Chtimes` after rewriting the file and asserts the change is still caught — it fails against the old check.

**Registrations in flight dropped (Copilot) — agreed, fixed.** The merge rebuilt the map purely from disk, so an account created by `AddUser` but not yet written was erased whenever an external edit was detected during its save. Records now carry a `persisted` flag, which distinguishes an account the editor deleted (drop it) from one this process created and has not written yet (keep it). That ambiguity was real and I had resolved it the wrong way.

**`nextUserID` after adoption (Major) — agreed, fixed.** Adopting a record with a higher ID now pushes the counter past it.

**Legacy handle on adoption (Minor) — agreed, fixed.** `Handle` is promoted from `LegacyUsername` exactly as `loadUsers` does, so the record is not written back handle-less and skipped by the next load.

**Test sleeps (Copilot) — resolved by the fingerprint change.** All six `time.Sleep(10ms)` calls are gone; change detection no longer touches the clock, so there is nothing to be flaky about.

**TOCTOU window (Major, heavy lift) — valid, filed as #216 rather than half-done.** Closing it needs an advisory lock held across read-merge-write and honoured by both processes, which means `flock`/`LockFileEx`/lock-file portability and teaching `./ue` to take the same lock. That is a cross-cutting change to two processes and both platforms. #216 also picks up the related non-atomic `os.WriteFile` in the same path, since both concern how the file is written and `./ue` already does it correctly.

Fourteen tests now, `go test ./... -race` clean. Verified live again: the sysop's demotion survives the login **and** the session's call count advances 15 → 16, which is the counter fix visible end to end.

PR #217, opened 2026-09-05
fix(reader): stop the screen bumping when a reader notice appears

### LINKED ISSUE #200: Don't bump the screen during Message Newscan and Message Reading
When you are doing a newscan or reading a message, selecting an option from the Reading menu inserts a row at the bottom that "bumps" the screen up, and it's jarring. Instead, replace the last row (the menu) with the message. That menu is no longer needed if the "Next" has been selected. or "No more messages" displays.

robbiew:
Closes #200.

## The bump, measured

Reproduced over telnet by driving the reader to the last message and pressing Next. The bytes sent for the notice were:

```text
b'\r\r\n\x1b[0;37m'   then   "End of messages."
```

No cursor positioning in front of them. The reader parks the cursor on the last row after drawing the lightbar (`message_reader.go:332`), and every one of these strings carries a leading `\r\n` for use in scrolling output:

```json
"msgEndOfMessages": "\r\n|07End of messages.|07"
```

Writing a newline on the bottom row makes the terminal scroll the whole screen up to make room. That is the jarring bump — the message body shifts under the reader's feet.

## The fix

Write the notice onto the lightbar row instead: position, erase the line, print without the leading break.

```text
b'\x1b[25;1H\x1b[2K\x1b[0;37m'   then   "End of messages."
```

This also matches what the issue asks for semantically — replacing the menu rather than appending below it. Once "End of messages" is showing the reader is leaving, so the options underneath are no longer selectable.

Applied to all five sites that print while the cursor is parked on that row: end-of-messages and already-at-first-message, each on both the filtered and unfiltered navigation paths, plus the deferred mail reply.

The strings themselves are untouched — they keep the leading break for scrolling output elsewhere, and it is stripped at the point of use.

## Testing

`go build ./...`, `go vet`, `go test ./...` clean. Unit tests cover the leading-break trim across all three shipped notice strings (including that trailing breaks are deliberately left alone), and that the erase sequence carries no cursor-positioning final byte.

Verified live on a real BBS for both navigation cases:

| | Before | After |
| --- | --- | --- |
| Next at last message | `\r\r\n` — scrolls | `ESC[25;1H ESC[2K` — replaces the row |
| Prev at first message | `\r\r\n` — scrolls | `ESC[25;1H ESC[2K` — replaces the row |

## Not covered

The issue also mentions newscan. `showScanNotice` (`message_scan.go:172`) writes `scanComplete` the same way, but it runs in a scrolling context rather than under a positioned lightbar, so it is not the same defect. If the newscan summary is also bumping for you, that is worth a separate look — say so and I will chase it.

### DIFF
--- a/internal/menu/message_reader.go
+++ b/internal/menu/message_reader.go
@@ -504,16 +504,14 @@ readerLoop:
 						currentMsgNum = nxt
 						break scrollLoop
 					}
-					terminalio.WriteProcessedBytes(terminal, ansi.ReplacePipeCodes([]byte(e.LoadedStrings.MsgEndOfMessages)), outputMode)
-					time.Sleep(500 * time.Millisecond)
+					showReaderNotice(terminal, outputMode, e.LoadedStrings.MsgEndOfMessages, termHeight)
 					break readerLoop
 				}
 				if currentMsgNum < totalMsgCount {
 					currentMsgNum++
 					break scrollLoop // Exit scroll loop to load next message
 				} else {
-					terminalio.WriteProcessedBytes(terminal, ansi.ReplacePipeCodes([]byte(e.LoadedStrings.MsgEndOfMessages)), outputMode)
-					time.Sleep(500 * time.Millisecond)
+					showReaderNotice(terminal, outputMode, e.LoadedStrings.MsgEndOfMessages, termHeight)
 					break readerLoop
 				}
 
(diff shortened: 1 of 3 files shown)


robbiew:
Both addressed in the latest commit — fair points, and both about the tests rather than the change itself.

**Inconsistent comment.** Correct, it could not mean both things. It claimed each notice string "must" begin with a line break in its stored form, then said the test still passes if one is rewritten without it. Reworded to state the property actually pinned: whatever the stored string looks like, the form written to the lightbar row must not open with a break, since that break is what scrolled the screen. Sysops can edit these strings via the string editor, so the point of use is the right place to pin it.

**Redundant assertion.** Also correct — once `eraseLine` is pinned to the exact value on the line above, the `ContainsAny(eraseLine, "Hf")` check can never fail, so it added no coverage while implying other cursor-final bytes were considered. Removed, and the test renamed to `TestEraseLineIsEraseInLine` to say what it actually verifies.

No change to the fix itself; `go test ./...` still clean.

Issue #203, opened 2026-09-04
Remove default doors from ANSI Doors Menu


robbiew:
When the BBS ships with default menu set, no doors are setup. The ANSI menu should not contain placeholder doors and commands.

robbiew:
Cross-reference for whoever picks this up: PR #218 does the equivalent cleanup for the **Main** menu, and the mechanism there turned out to be different from what this issue describes.

MAIN's dead entries were `RUN:PLACEHOLDER` bindings — a handler that prints the same "undefined input" as an unbound key. Scanning the whole menu set, **all six placeholder bindings were in MAIN**; `DOORSM.CFG` has none. So the Doors problem is not stale bindings but art advertising doors that are not configured in `doors.json`, which needs a different fix.

Useful precedent from #218 if the art needs editing: the file carries a SAUCE trailer whose `FileSize` field should be updated to match the edited body (the BBS strips SAUCE on load and never reads it, but other ANSI tools do), and column separators want checking afterwards — MAIN's land on columns 2, 22, 42, 61 on every row.

PR #218, opened 2026-09-06
fix(menus): correct and clean up the default Main menu

### LINKED ISSUE #202: Default Main menu has duplicate and missing items
The ANSI main menu needs to be updated to include all available commands, plus fix the duplicate "W" item

robbiew:
Closes #202.

Audited every entry in `MAIN.ANS` against `MAIN.CFG` in both directions, then followed up on what the audit turned up.

## 1. The duplicate W and the missing command are one defect

The art offered both `(W) Last Callers` and `(W) Who's Online`, but only one `W` is bound — to Last Callers. **Who's Online is `/W`**, so the second entry was unreachable and pressing `W` silently gave the wrong screen. Verified live before fixing.

Checking the other direction, everything bound but not advertised was a placeholder or an alias of something already shown — except `/W`, the very command being mislabelled. So the issue's two halves were one defect seen from either side.

Corrected to `(/W) Who's Online`, paying for the extra character with one fewer trailing space so the columns do not move.

## 2. Placeholders retired

Six entries ran `RUN:PLACEHOLDER`, which prints the same "undefined input" as an unbound key. They split into two groups deserving opposite treatment.

**Three were stubs for features that already exist**, reachable elsewhere but dead on MAIN:

| Key | Was | Now | Also reachable from |
| --- | --- | --- | --- |
| `U` | `PLACEHOLDER AutoSig` | `RUN:CFG_AUTOSIG` | MSGMENU `S`, USERCFG `R` |
| `+` | `PLACEHOLDER ChangePW` | `RUN:CFG_PASSWORD` | USERCFG `P` |
| `Z` | `PLACEHOLDER PointerSet` | `RUN:UPDATENEWSCAN` | MSGMENU `U` |

Verified live — each now reaches its real screen:

```text
(U) -> "Auto-Signature ... Change/create  Delete  Quit :"
(+) -> "Current Password:"
(Z) -> "Set NewScan Date; All New, None (Mark All Read), or Enter Date"
```

**Three had no implementation anywhere** — `TopTenStats`, `FEEDM`, `ChangeMenu` — and are removed. Pressing them is unchanged, since a placeholder and an unbound key behave identically; what changes is that the config no longer implies the features exist. `MAIN.CFG` now has no placeholder entries at all.

`(F) Feedback` was also the one placeholder the art advertised, so it is gone from the art too — following the principle in #203 that shipped menus should not advertise placeholders.

## 3. The Sysop row used the wrong separator

A literal `|` (0x7C) where all other rows use `│` (0xB3), so it rendered as `|(%) Sysop Menu`. Corrected.

## Result

```text
│(M) Messages Menu  │(L) User List      │(O) One-Liners    │(^) Version
│(T) Transfer Menu  │(W) Last Callers   │(H) User Voting   │(S) System Stats
│(D) Doors Menu     │(/W) Who's Online  │(V) Vote Booth    │(U) Autosig
│(C) Chat           │(B) BBS Listings   │(I) Infoforms     │(G) Logoff
│(J) News           │(Y) Your Stats     │(P) Page User     │(/G) GTFO!
│(E) Private Email  │(K) Your Konfig    │(R) Rumors        │(+) Change Pass
│(Z) Newscan Ptrs
│(%) Sysop Menu
```

Reviving three commands takes the menu to 25 entries, which does not fit the four-by-six grid. The last one uses the blank line that already separated the commands from the sysop row, so the block gains a row without the art growing at all. Labels follow the wording already on MSGMENU ("Autosig"). All four separators still land on columns 2, 22, 42, 61; widest line is 78.

**SAUCE `FileSize`** is kept accurate rather than left stale — the BBS strips SAUCE and never reads it, but other ANSI tools do.

## Testing

`go build ./...` and `go test ./...` clean. Verified live throughout: the mislabelled `W`, the dead `F`, each of the three revived commands, and that the three dropped keys are inert.

## Summary by CodeRabbit

- **New Features**
  - Enabled menu commands for viewing top-ten statistics and feeds.
  - Added menu actions for editing autosignatures, updating pointers, and changing passwords.

- **Removed**
  - Removed the placeholder entry for changing menu sets.


robbiew:
Went with the second option — advertise `G`, hide `/G` — rather than combining them, because they are not the same action:

- **`G`** (`runMainLogoffCommand`) prompts *"Log off now?"* and can be backed out of
- **`/G`** (`runImmediateLogoffCommand`) hangs up immediately, no confirmation

A shared label like `(G /G) GTFO` would mean someone pressing `/G` expecting the prompt just gets dropped. Not a trade worth a menu slot.

Hiding it costs nothing: `/G` stays bound and working, it is already documented at `docs/sysop/menus/menu-system.md:212` ("Hang up immediately, no confirmation"), and MAIN already keeps four other shortcuts off the art the same way — `N`, `X`, `!`, `/SYSOP`.

That frees the slot Newscan Ptrs needed, so the ragged seventh row is gone and the block is a clean six rows of four:

```text
│(M) Messages Menu  │(L) User List      │(O) One-Liners    │(^) Version
│(T) Transfer Menu  │(W) Last Callers   │(H) User Voting   │(S) System Stats
│(D) Doors Menu     │(/W) Who's Online  │(V) Vote Booth    │(U) Autosig
│(C) Chat           │(B) BBS Listings   │(I) Infoforms     │(G) Logoff
│(J) News           │(Y) Your Stats     │(P) Page User     │(Z) Newscan Ptrs
│(E) Private Email  │(K) Your Konfig    │(R) Rumors        │(+) Change Pass

│(%) Sysop Menu
```

24 entries, every row full, one separator layout (columns 2/22/42/61), widest line 78. The blank separator before the sysop row is back.

Verified live: the menu renders as above, and `/G` still shows GOODBYE.ANS and hangs up with no prompt.

PR #219, opened 2026-09-06
feat(scripting): expose previousLogin to scripts

### LINKED ISSUE #208: Expose previousLogin to JS scripts alongside lastLogin
Follow-up from #198 / PR #205.

`userToJS` in `internal/scripting/users.go:73` exposes `lastLogin` to scripts:

```go
jsutil.Set(obj, "lastLogin", u.LastLogin.Unix())
```

`UserMgr.Authenticate` stamps `LastLogin = time.Now()` at authentication, *before* the login sequence and any script runs. So by the time a script reads `user.lastLogin`, it is the current session's timestamp, not the caller's previous visit.

Any script written to ask "what is new since this user was last here?" — a custom newscan, a bulletin gate, a welcome-back banner — silently gets an empty result, with nothing to indicate why. This is the same trap that made `PRINTNEWS` never display news, and that made the `|LCALL` placeholder always render today's date; both were fixed in PR #205.

PR #205 adds `User.PreviousLogin`, which holds the prior stamp, but does not expose it to the scripting bridge.

## Proposed change

Add the counterpart field:

```go
jsutil.Set(obj, "previousLogin", u.PreviousLogin.Unix())
```

Points to settle:

- **Zero value.** A first-time caller has a zero `PreviousLogin`, and `time.Time{}.Unix()` is `-62135596800`, which is a confusing thing to hand a script. Consider exposing `0`, or `null`, for "no previous visit" — and whatever is chosen, document it.
- **Leave `lastLogin` as-is.** It is accurate for what it says and scripts may depend on it. Documenting the distinction is enough.
- **Docs.** Update the scripting API reference so the difference between the two is explicit, since the failure mode is silent.

Deliberately left out of PR #205: it is an API surface addition rather than a fix, so it belongs on its own.

robbiew:
Closes #208.

`userToJS` handed scripts `lastLogin` and nothing else time-related. That stamp is written at authentication — before the login sequence and before any script runs — so a script asking *"what is new since this user was last here?"* compares against the current session and silently finds nothing.

It is the same trap that stopped system news displaying (#198), made the rumors and file newscans find nothing, and had `|LCALL` rendering today's date. All fixed in the BBS itself; the scripting bridge was the last place it survived.

## The zero value

A caller with no previous visit reports **`0`**, not `time.Time{}.Unix()` which is `-62135596800`.

Zero is falsy, so `if (user.previousLogin)` reads naturally, and it is early enough that `postedAt > user.previousLogin` treats everything as new — the right answer for a first-time caller. The raw negative behaves identically in a numeric comparison but is baffling to print or log.

## `lastLogin` is deliberately unchanged

It is accurate for what it says, and scripts do depend on it. The shipped `scripts/examples/last10.js` sorts users by `lastLogin` to find the most recent callers — a legitimate use that has nothing to do with the per-user "since your last visit" question. Changing its meaning would break it.

The reference now documents the distinction with both forms side by side, since the failure mode is an empty result rather than an error:

```javascript
// Right: finds everything posted since the caller was last on.
var fresh = items.filter(function (i) { return i.postedAt > user.previousLogin; });

// Wrong: lastLogin is this session, so this is always empty.
var none = items.filter(function (i) { return i.postedAt > user.lastLogin; });
```

## Testing

`go build ./...`, `go vet`, `go test ./...` clean. Four tests run against a real goja VM through `userToJS`, covering the zero-time mapping, both fields being exposed and distinct, a first-time caller getting a falsy `0`, and — the one that matters — a script filtering a list of timestamps, asserting `previousLogin` finds the 2 items posted since the last visit while `lastLogin` finds 0.

That last test fails against the old behaviour, which is the point.

## Summary by CodeRabbit

- **New Features**
  - Scripting user objects now include `previousLogin`, a Unix timestamp for the user’s prior visit.
  - First-time users receive `0`, allowing scripts to identify users who have not logged in previously.
  - Scripts can compare `previousLogin` with `lastLogin` to identify content added since a user’s last visit.

- **Documentation**
  - Clarified the difference between `lastLogin` (current session) and `previousLogin` (previous visit).
  - Added examples for accurately scanning newly added content.

### DIFF
--- a/docs/sysop/scripting/vpl-scripting.md
+++ b/docs/sysop/scripting/vpl-scripting.md
@@ -360,13 +360,34 @@ Read-only access to the user database. Returns safe fields only — no passwords
 | `filePoints` | number |
 | `validated` | boolean |
 | `lastLogin` | number (Unix timestamp) |
+| `previousLogin` | number (Unix timestamp, `0` if never) |
 | `createdAt` | number (Unix timestamp) |
 
-> **`lastLogin` is the current session, not the caller's previous visit.** It is
-> stamped at authentication, before any script runs, so a script comparing
-> against it to find "what is new since this user was last here" will always
-> come up empty. The user record's `previousLogin` field holds the prior visit
-> but is not yet exposed to scripts — see issue #208.
+#### `lastLogin` vs `previousLogin`
+
+`lastLogin` is **the current session**, not the caller's previous visit. It is
+stamped at authentication, before any script runs, so it always reads as "a
+moment ago".
+
+To ask *"what is new since this user was last here?"* — a custom newscan, a
+bulletin gate, a welcome-back banner — compare against `previousLogin`:
+
+```javascript
+// Right: finds everything posted since the caller was last on.
+var fresh = items.filter(function (i) { return i.postedAt > user.previousLogin; });
+
+// Wrong: lastLogin is this session, so this is always empty.
+var none = items.filter(function (i) { return i.postedAt > user.lastLogin; });
+```
+
+The failure is silent — the second form returns an empty list rather than an
+error — which is what made the same mistake go unnoticed in the BBS itself for
+system news, the rumors newscan, and the `|LCALL` display placeholder.
+
+`previousLogin` is **`0`** for a caller with no previous visit. That is falsy,
+so `if (user.previousLogin)` reads naturally, and it is early enough that a
+`>` comparison treats everything as new — the right answer for a first-time
+caller.
 
 ---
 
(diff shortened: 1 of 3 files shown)


robbiew:
Good catch — fixed.

The assertion is now the comma-ok form and reports what actually came back, plus a length check before indexing:

```go
exported := evalWithUser(t, u, expr).Export()
res, ok := exported.([]interface{})
if !ok {
    t.Fatalf("script returned %T (%v), want a two-element array", exported, exported)
}
```

`toInt` had a quieter version of the same problem: it returned `-1` for an unexpected type, which would have surfaced as a confusing count mismatch ("found -1 new items, want 2") rather than as "the script did not return a number". It now fails with the offending type and value.

PR #220, opened 2026-09-06
docs(users): correct the new-user defaults and state what signup sets

### LINKED ISSUE #204: Clarify in docs that default settings for new users is "10" and non-validated
Many sysops will auto-validation and set the user to "regular" user so they can do all the things from first login. Docs should clarify that the default state is non-validated

robbiew:
Closes #204.

The issue asks for the new-user defaults to be documented. Checking them against the source turned up that most of the surrounding numbers were wrong, because **two different things are called "defaults" and the docs mixed them.**

## Two sets of defaults

`templates/configs/config.json` is what a normal install runs — setup copies it into `configs/`. The built-in fallback only applies to keys *absent* from that file: `LoadServerConfig` starts from `defaultConfig` and unmarshals the file over it.

Five settings differ:

| Setting | Shipped | Fallback |
| --- | --- | --- |
| `invisibleLevel` | `250` | `0` |
| `newUserLevel` | `10` | `1` |
| `regularUserLevel` | `25` | `10` |
| `anonymousLevel` | `50` | `5` |
| `deletedUserRetentionDays` | `-1` | `30` |

`configuration.md` documented the **fallback** column throughout — the set almost nobody experiences. It now shows both, labelled.

`user-management.md` was worse: its sample config block showed the fallback values while the prose beside it described different ones, and its `regularUserLevel` figure of **20 matched neither**. Corrected against the shipped config, along with the worked example and the tiering walkthrough that both depended on 20.

The alternative tiering examples are deliberately left alone — they are different configurations offered as options, not claims about defaults.

## The issue's actual ask

A new section states what signup produces:

| Field | Value |
| --- | --- |
| `accessLevel` | `newUserLevel` — `10` shipped |
| `validated` | **`false`**, always |

The part worth spelling out is that **`validated: false` is not configurable** — it is set unconditionally at account creation, so a sysop looking for an auto-validation switch will not find one. The issue notes many sysops want exactly that, so the docs now name the three actual levers: raise `newUserLevel`, validate by hand, or enable New User Voting.

Also notes that level and validation move independently, so raising a level does not satisfy an ACS check written against validation status.

## Verification

Rather than eyeballing, I checked the documented numbers programmatically against the source of truth:

```text
sample block vs shipped config: MATCHES
configuration.md shipped column: MATCHES
stale 'regularUserLevel (20)': gone
stale 'anonymousLevel (5)': gone
stale 'newUserLevel ... (default: `1`)': gone
```

Docs-only change; `go build ./...` and `go test ./...` still clean.

### DIFF
--- a/docs/sysop/configuration/configuration.md
+++ b/docs/sysop/configuration/configuration.md
@@ -446,7 +446,7 @@ built-in defaults and overlays whatever the file provides.
 | `logonLevel` | `10` | `10` | Minimum level required to log in |
 | `anonymousLevel` | `50` | `5` | Minimum level to post anonymously (`0` = disabled) |
 
-See [User Management](../users/user-management.md#system-access-levels) for how
+See [User Management](users/user-management.md#system-access-levels) for how
 these interact, and what happens at signup.
 
 **Default Settings:**
--- a/docs/sysop/configuration/configuration.md
+++ b/docs/sysop/configuration/configuration.md
@@ -428,19 +428,30 @@ General BBS configuration. All settings in this file are managed through the **S
 
 **Access Levels:**
 
-- `sysOpLevel` — Security level for SysOp access (default: `255`)
-- `coSysOpLevel` — Security level for Co-SysOp access (default: `250`)
-- `invisibleLevel` — Level at which a user is invisible in the who's-online list (default: `0`, falls back to `coSysOpLevel`)
-- `newUserLevel` — Level assigned to a brand-new account (default: `1`)
-- `regularUserLevel` — Level for validated/regular users (default: `10`)
-- `logonLevel` — Level granted on successful login (default: `10`)
-- `anonymousLevel` — Level for guest/anonymous access (default: `5`, `0` = disabled)
+Two different sets of values are called "defaults" here, and they differ. The
+**shipped** value is what a normal install gets, since setup copies
+`templates/configs/config.json` into `configs/`. The **fallback** applies only
+when a key is absent from your `config.json` — the loader starts from the
+built-in defaults and overlays whatever the file provides.
+
+| Setting | Shipped | Fallback | Meaning |
+| ------- | ------- | -------- | ------- |
+| `sysOpLevel` | `255` | `255` | SysOp access |
+| `coSysOpLevel` | `250` | `250` | Co-SysOp access |
+| `invisibleLevel` | `250` | `0` | Invisible in who's-online (`0` falls back to `coSysOpLevel`) |
+| `newUserLevel` | `10` | `1` | Assigned to a brand-new account |
+| `regularUserLevel` | `25` | `10` | Assigned when a user is validated |
+| `logonLevel` | `10` | `10` | Minimum level required to log in |
+| `anonymousLevel` | `50` | `5` | Minimum level to post anonymously (`0` = disabled) |
+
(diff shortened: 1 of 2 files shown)


robbiew:
All three addressed — thanks, the `invisibleLevel` one in particular.

**Broken anchor.** Correct: `**Access Levels:**` is a bold label, not a heading, so `#access-levels` resolved to nothing. Its siblings in that section (`**Server Setup:**`, `**Connection Limits:**`, `**Default Settings:**` …) are all bold labels too, so promoting just this one to a heading would look inconsistent. Added an explicit `<a id="access-levels">` instead, which keeps the style and makes the link stable.

**`invisibleLevel` description.** Verified in the code rather than taking it on faith, and you are right — I had inherited the wrong description from the old text. `cmd/vision3/main.go` uses it as the threshold for *offering* the Invisible Logon prompt; the caller's answer then decides whether the session is added to the Last Callers list. It does not by itself mark anyone invisible. Reworded to match.

**Missing `invisibleLevel` in the fallback list.** Also right, and it undercut the point the paragraph was making — the whole passage is about shipped-versus-fallback drift, so listing three of the four differences was the wrong place to be incomplete. Now says four and names them.

Re-ran the programmatic check afterwards: the shipped column still matches `templates/configs/config.json` exactly, the anchor is present, and the fallback list names all four.

robbiew:
Pushed `d8fe4eb` with @robbiew's fixes, plus two follow-ons found while checking them.

**The link fixes are right and mine were broken.** I used filesystem-relative paths (`../users/…`, `../configuration/…`), but this tree is served by docsify, which resolves from the docs root — and every pre-existing link follows that form (`users/admin-menu.md` from *inside* `users/`, `menus/menu-system.md` from `menus/`). Both of my links pointed at nothing.

Checking that turned up a third one I had added with the same fault: `[New User Voting](nuv.md)` needed `users/nuv.md`. All 23 cross-file links across the two pages now resolve.

**The `accessLevel: 1` corrections are also right**, and in a section I had not touched. `AddUser` assigns `um.newUserLevel` — `10` shipped — and `1` is only the fallback. Same shipped-versus-fallback confusion this PR exists to fix.

**"Validated users" → "Users at level 25+"** is a good call for the same reason: the page's own argument is that access follows level rather than the flag, so labelling a level band by validation status undercut it.

### One more, pre-existing

The new-account summary further down said:

> `validated: false` — user cannot log in until a SysOp sets this to `true`

which directly contradicts the Login Flow section above it ("does **not** block login"). Verified in code — `executor_login_auth.go:177` gates on `AccessLevel < LogonLevel`, and nothing reads `Validated`. Since this PR is specifically about what signup produces, leaving that contradiction in place seemed wrong, so it now points at the `validated` flag section instead.

robbiew:
Filed #222 for the behavioural side of this.

The contradiction I corrected here turned out to be the visible edge of a wider mismatch: `internal/menu/newuser.go:247` tells **every** new signup "requires SysOp validation, please call back later" with no config check around it, and with the shipped defaults (`newUserLevel` 10, `logonLevel` 10) that caller can log straight in. The matrix "check your access" path at `matrix.go:283` does the same.

So this PR makes the documentation honest while the BBS still is not. Kept separate deliberately — #220 is a docs change and should not grow a behaviour change.

PR #225, opened 2026-09-06
feat(messages): optional groups in message header templates

### LINKED ISSUE #209: Templates need a way to drop decoration around an empty placeholder

robbiew:
Closes #209. Also resolves item 3 of #201.

Option 1 per the decision on #209 — hide the row when empty, leave `@U@`'s meaning alone.

## The semantic difference from prompts

Prompts got `|{ ... |}` in #210. Headers need the same syntax but not the same behaviour.

A prompt is flowing text: dropping a group removes it and the line closes up. A header is not — it draws boxes, pads values to fixed columns, and four of the fifteen templates position the cursor absolutely. Removing characters would pull a border out of line.

**So in a header, a dropped group is replaced by spaces occupying the width it would have rendered at.** Nothing moves; the decoration goes blank.

That was the open question on this issue, and the templates answered it: `@U@` is mostly *inline* — bracketed, or beside another field, sometimes inside a box border — not a standalone row that could simply be removed.

Width accounting follows the templates' own convention: a `#` token reserves the length of the **whole token** (`@U#####@` is eight characters, reserves eight columns), which is what makes padding line up in an ANSI editor. Bare `@U@` collapses to nothing. ANSI escapes count zero.

## What this fixes

Nine templates wrapped. The artifacts are precisely what #201 opened with:

| Template | Rendered with no note |
| --- | --- |
| `MSGHDR.4`, `.7` | literal **`()`** |
| `MSGHDR.3` | **`[]`** |
| `MSGHDR.2` | **`""`** |
| `MSGHDR.12` | stray unclosed **`[`** mid-line |
| `MSGHDR.10`, `.15` | stranded `User Note:` / `Note:` label |
| `MSGHDR.1`, `.9` | gap between adjacent fields |

## MSGHDR.8 is deliberately not wrapped

Its `Note..:` label is drawn in the static box art and the value is placed separately by absolute cursor positioning, so **a contiguous group cannot cover both**. Hiding the value alone would strand the label — the problem rather than the fix. Documented as a limitation of contiguous groups rather than worked around.

## Testing

`go build ./...`, `go vet`, `go test ./...` clean. Seven unit tests cover blanking on empty, keeping when populated, padded-width preservation, that following content does not shift, escapes not inflating the width, whitespace counting as empty, and the edge cases (no placeholder, mixed values, unknown code, unmatched opener).

Verified live across styles 3, 4, 7, 10, 12 and 15 with an empty note — no `()`, `[]` or `""` in any of them. Checked the other direction too, so this blanks on emptiness rather than unconditionally:

```text
empty note -> " Name: @F@   "
note set   -> " Name: @F@ [@U@]"
```

## Note on `@U@` itself

This hides the row; it does not change what `@U@` means. Worth remembering from the V/2 source that `|U` was `B.Status`, a **per-message** field, and V/3 repointed it to the author's SysOp-only `privateNote` — which is why it is blank for essentially everyone. Repointing it remains open as a separate product question; this PR takes option 1 as decided.

## Summary by CodeRabbit

- **New Features**
  - Message header templates now support optional groups that hide empty values while preserving column alignment.
  - Header spacing correctly accounts for padding and formatting codes.
  - Populated values continue to display with their configured formatting.

- **Bug Fixes**
  - Improved handling of empty or whitespace-only header values without shifting surrounding content.
  - Updated several message header layouts for more consistent conditional user-field formatting.

- **Documentation**
  - Added guidance for using optional groups in message header templates, including current layout limitations.


robbiew:
Fixed — and the finding is real for a sharper reason than the example.

`@Q@` does not parse: `placeholderRegex` has a fixed code set (`[BTFSUL#NDWPEOMAZCXGVK]`), so an unrecognised letter is literal text and the group is already kept. There is a test for that case.

**`@G@` is the one that bites.** It parses, but is deliberately absent from the substitution map because it is gap fill, expanded later against the terminal width. Skipping it meant a group holding a gap fill beside an empty value was blanked outright:

```text
"|{@U@@G@|}"  with an empty @U@  ->  ""     (gap fill destroyed)
```

Reproduced before fixing. An unevaluatable placeholder now keeps the group — "we cannot tell what this renders as" is not a basis for deciding it renders as nothing.

Checked the nine wrapped templates do not regress: every group in them contains only `@U@`, verified programmatically rather than by eye, and blanking still works live with an empty note. `MSGHDR.9` does use `@G*@`, though not inside a group — but a template author wrapping a row that contains one would have hit this.

robbiew:
Missed the nitpick on the first pass — I read the inline comments via the API and acted on those, but this one lives in the review *body* rather than as an inline comment, so it did not show up. Fixed now.

Two markdown warnings, both in the section this PR adds:

- the fenced example had no language — now `text`
- the sentence after it used an inline code span with spaces immediately inside the backticks (MD038)

The spaces were deliberate: they showed the trailing blanks the padding leaves behind. But an inline span is the wrong place to make whitespace visible — it renders as ordinary text, so the point was lost anyway and the warning is fair. Both states now appear in a fenced block, where the alignment is actually legible:

```text
note set     Note: sysop
note empty
```

Pre-existing untagged fences elsewhere in the file are left alone, since the review scopes to this diff and sweeping them would bury the change.

PR #226, opened 2026-09-06
feat(users): add autoValidateNewUsers, and stop the BBS misdescribing access

### LINKED ISSUE #222: New user is told to wait for validation, then can log straight in
Found while fixing the documentation in #220 / PR #220. The docs contradiction there turned out to be the visible edge of a mismatch between what the BBS tells users and what it does.

## The problem

Access is decided by level alone. `internal/menu/executor_login_auth.go:177` gates on `AccessLevel < LogonLevel`; nothing anywhere reads `Validated`. The flag is administrative — validating a user triggers an upgrade to `regularUserLevel` and marks them reviewed, but it does not gate anything.

The shipped strings say otherwise, and they are shown **unconditionally**:

`internal/menu/newuser.go:247`, after every signup, with no config check around it:

> Your account has been created but requires **SysOp validation**.
> Please call back later to check your access.

`internal/menu/matrix.go:283`, on the "check your access" option:

> Account '%s' is **not yet validated**. Please wait for SysOp approval.

With the shipped defaults — `newUserLevel` 10, `logonLevel` 10 — a new caller is told to call back later for approval, and can then log in immediately with everything ACS `s10` allows.

## Why this matters

It is a misconfiguration risk rather than a cosmetic one. A sysop who believes signups are held pending review has no reason to check, because the board tells the *user* exactly that. Meanwhile every new account has live access. The user-facing message actively reinforces the wrong mental model, so nothing prompts anyone to discover the gap.

The documentation said the same thing until #220 corrected it, which is presumably where the expectation came from.

## Proposed fix

Both sites already have what they need to tell the truth. Make the message reflect the actual configuration:

- **`newUserLevel >= logonLevel`** — the account can log in now. Say so, and note that a sysop will review it.
- **`newUserLevel < logonLevel`** — the account genuinely cannot log in yet. The current wording is correct and should stay.

Same treatment for the matrix "check your access" path, which should report whether the account can actually log in rather than only whether it is validated.

This changes no access semantics — only what is said.

## What this deliberately does not propose

Adding a `requireValidation` setting to make `validated` gate login.

It is the obvious way to make the existing strings true, but it would create two mechanisms for one intent: `newUserLevel < logonLevel` already holds new accounts out, and that is what the docs now teach. A second lever means sysops have to reason about how the two interact, which is how the current confusion arose. Keeping `validated` unambiguously administrative, with levels as the single access mechanism, seems better.

Worth an explicit decision either way, since the counter-argument — that sysops expect the flag to be a gate strongly enough that levels alone will keep surprising them — is reasonable.

## Related

- #220 corrected the documentation, including a pre-existing line claiming `validated: false` blocks login.
- New User Voting (`useNuv`) is a third path here: when enabled it validates on a successful vote (`internal/menu/nuv.go:208`), so the same message applies to candidates whose access does not depend on the outcome.

robbiew:
Closes #222.

## 1. The setting

`AddUser` hardcoded `Validated: false`, so there was no way to have signups arrive reviewed. `autoValidateNewUsers` now controls it, **defaulting off** so an upgrading system behaves exactly as before. Plumbed like `newUserLevel` — set from config at startup and again on reload, so it takes effect without a restart.

Worth being precise about what it buys, because the name promises more than it delivers. **Validation does not gate login** — access level against `logonLevel` does — so this does not decide whether a new caller gets on. It marks the account reviewed, which suppresses the required-infoforms prompt (`infoforms_commands.go:158`) and clears it from the pending list.

## 2. Mutually exclusive with New User Voting

Per the decision on the issue, enforced rather than warned about.

The config editor refuses to enable either while the other is on, **in both directions**, naming the setting that is in the way:

```text
cannot auto-validate while New User Voting is enabled; turn Use NUV off first
cannot enable New User Voting while Auto Validate is on; turn Auto Validate off first
```

Turning either **off** is always allowed — otherwise a sysop could not escape the state to switch to the other one. There is a test for that specifically.

The editor only guards the UI, so `LoadServerConfig` enforces it too for a hand-edited `config.json`. **Auto-validation is the one that loses**: keeping the review process is the safer outcome, and silently validating every signup is not something to fall into by way of a typo. It logs and carries on rather than refusing to boot:

```text
ERROR autoValidateNewUsers and useNuv are mutually exclusive; disabling
      autoValidateNewUsers for this run
      path=/Users/robbie/bbs-dev/configs/config.json
      fix="set one of them to false in config.json"
```

## 3. The messages

`newuser.go:247` told **every** signup *"requires SysOp validation, please call back later"* with no condition attached, and `matrix.go`'s "check your access" branched on the validated flag. Both described a gate that does not exist — with the shipped defaults (`newUserLevel` 10, `logonLevel` 10) a caller was told to wait and could then log straight in.

Both now ask `canLogonAtLevel`, which mirrors what `executor_login_auth.go:177` actually enforces. Signup gains a `newUserAccountReady` string for the case the BBS previously had no way to express, and the existing strings are reworded to talk about access rather than validation. The matrix message now reports the caller's level and the minimum, so "cannot log on yet" says why.

## Testing

`go build ./...`, `go vet ./...`, `go test ./...` clean.

Twelve tests: the logon predicate across thresholds (including that it ignores validation entirely, which is the distinction this issue is about), `AddUser` honouring the setting in both states and responding to a reload, load-time conflict resolution, each setting alone being preserved, the default staying off, and the editor guards in both directions plus the turn-off escape.

Verified live: with both settings enabled in `config.json`, the BBS starts, logs the conflict, and runs with auto-validation disabled.

## Summary by CodeRabbit

* **New Features**
  * Added an option to automatically validate newly registered accounts.
  * New-user messages now distinguish immediate access, approval requirements, and pending review.
  * Login eligibility now reflects configured access levels independently of validation status.
  * Matrix access responses identify when an account does not meet the required access level.

* **Bug Fixes**
  * Prevented automatic validation and New User Voting from being enabled together.
  * Configuration reloads now apply updated automatic-validation settings.


robbiew:
Signup message now verified live in both directions — this was the gap I flagged when opening the PR.

**Account can log on** (`newUserLevel` 10 ≥ `logonLevel` 10):

```text
Your account has been created. You can log on now.
```

**Account cannot** (`newUserLevel` 1 < `logonLevel` 10):

```text
Your account has been created, but a SysOp must raise your access
before you can log on.
Please call back later.
```

And the first message is true, which is the whole point of the issue. The account it created came out **level 10, `validated: false`** — and logged straight in on the next connection. That is precisely the case the old unconditional message told to call back later.

Note the blocked message says *"a SysOp must raise your access"* rather than *"requires validation"*. Raising the level is what actually unblocks them; validating alone would not, unless it happens to trigger the `regularUserLevel` upgrade past the threshold. Saying "validation" there would have reproduced the original bug in the opposite direction.

Test accounts and the temporary `newUserLevel` change have been cleaned up.

robbiew:
Both findings fixed in `196d2e3`, and the first led me to a worse one I had introduced.

**Missing default for the new key — correct.** `applyStringDefaults` exists precisely so keys added later keep working against a `strings.json` that predates them, and `newUserAccountReady` was not in it. An upgraded install would have taken the ready path and printed **an empty line** — silence, where the old code at least said something wrong. Added, along with defaults for the other two new keys, plus a test that loads a `strings.json` containing none of them and asserts each still renders.

**The one that finding exposed.** `matrixAccountNotValidated` is a **user-editable format string**, and I had changed it from one verb to three while passing three arguments. Every `strings.json` already in the wild still holds the one-verb text, so the reply would have rendered:

```text
Account 'NewGuy' is not yet validated. Please wait for SysOp approval.
%!(EXTRA int=1, int=10)
```

Reproduced before fixing. The replacement is now a separate key, `matrixAccountCannotLogon`, and the original is restored untouched and no longer read — an install that customised it does not get a silent change of meaning. There is a test pinning the two arities apart so this cannot drift back.

**Review note in the ready message — agreed.** "You can log on now" alone implied nothing further happens. The signup message now appends `newUserPendingReview` when `Validated` is false, which also means it correctly stays silent when `autoValidateNewUsers` has already marked the account reviewed.

Verified live end to end:

```text
Your account has been created. You can log on now.
A SysOp will review your account.
```

Test accounts cleaned up afterwards.

PR #228, opened 2026-09-07
feat(users): continue into the session after signup

### LINKED ISSUE #227: After signup, drop the new user into the session instead of back at the matrix
Raised by robbiew while reviewing #226.

## Current behaviour

A caller creates an account — chooses a handle, sets a password, fills in the details — and is then returned to the matrix menu, where they must select "Journey onward" and type the handle and password they just chose.

`internal/menu/matrix.go:236`:

```go
return "MATRIX", nil // Return to matrix after signup
```

Confirmed live: immediately after the signup message the screen shows `Journey onward / Create an account / Check your access / Disconnect`.

It is not broken — "Journey onward" is right there — but it makes the caller re-enter credentials they set thirty seconds earlier, and #226 sharpens the oddity by ending signup with "You can log on now". They can; they just have to do it themselves.

## Why this is not a small change

Nothing in the signup path carries the created user back out:

| Function | Returns today |
| --- | --- |
| `handleNewUserApplication` | `error` only — the created user is discarded |
| `RunMatrixScreen` | a string action only |
| matrix `case "NEWUSER"` | the literal `"MATRIX"` |

Continuing into the session means threading the user through all three and teaching `main.go`'s login loop to accept a pre-authenticated caller rather than always prompting. That is a change to the matrix API and the session entry path, which is why it is filed separately rather than folded into #226.

## Points to settle

- **When should it apply?** The natural condition is `canLogonAtLevel` — the same predicate #226 added — so a caller whose level is below `logonLevel` still gets told to call back, since continuing would only bounce them. Validation should not enter into it, for the reasons #222 covers.
- **Should it be optional?** Some sysops may want a fresh login as a deliberate confirmation that the password was memorised. Worth considering, though #222 already argued against settings that duplicate an existing mechanism, and this one has no existing equivalent to duplicate.
- **What about the login sequence?** A caller continuing straight from signup should presumably still get the full login sequence (news, last callers, and so on) rather than landing on a bare MAIN, so the entry point matters: it should join where a normal login joins, not after it.
- **`timesCalled` and `lastLogin`** are stamped by `Authenticate`, which this path bypasses. Whatever continues the session needs to set them, or the first call will not be recorded.

## Related

#226 makes the signup message describe reality. This makes the flow match the message.

robbiew:
Closes #227.

A caller who had just chosen a handle and password was returned to the matrix menu and made to type them again. #226 sharpened the oddity by ending signup with *"You can log on now"* — they could, but had to do it themselves.

## What changed

They now continue straight into the session, gated on `canLogonAtLevel`. An account whose level is below `logonLevel` still returns to the matrix, since continuing would only bounce it back out. Validation is deliberately **not** consulted — it does not gate login, and letting it gate this would reintroduce exactly the confusion #222 removed.

The plumbing did not carry the account out of signup, so three returns widened:

| | Was | Now |
| --- | --- | --- |
| `handleNewUserApplication` | `error` | `(*user.User, error)` |
| `processMatrixAction` | `(string, error)` | `(string, *user.User, error)` |
| `RunMatrixScreen` | `(string, error)` | `(string, *user.User, error)` |

`main.go` treats a user coming back from the matrix as authenticated and skips the login loop. Everything *after* that loop — the login sequence included — runs exactly as for any other caller, which was one of the points to settle on the issue. Confirmed by the required-infoform prompt appearing immediately after signup.

The two "type NEW at the login prompt" paths had the same gap and are fixed alongside. Simpler there, since both already return the authenticated user.

## The bookkeeping, which is the part worth care

`lastLogin`, `previousLogin` and `timesCalled` are stamped by `Authenticate`, which this path bypasses entirely — so a first call would have gone unrecorded, silently. I flagged this on the issue as the easy thing to miss.

Rather than repeat the stamping, `Authenticate`'s half is extracted into `UserMgr.BeginSession` and **both paths call it**, so they cannot drift. `Authenticate` now verifies the password and delegates.

## Testing

`go build ./...`, `go vet ./...`, `go test ./...` clean. Four tests on `BeginSession`: a first call recorded with `PreviousLogin` left zero, a later call rolling the stamp, the returned snapshot matching what was stored, and an unknown handle. Existing `Authenticate` tests pass unchanged against the refactor, which is the point of extracting rather than duplicating.

Verified live, both directions:

```text
level 10 (>= logonLevel)  -> lands in the BBS
                             timesCalled=1, lastLogin set, previousLogin zero
level 1  (<  logonLevel)  -> returns to the matrix, timesCalled=0
```

`previousLogin` being zero is what makes the newscans treat everything as new for a first-time caller — the behaviour established in #205 and #214.

## Summary by CodeRabbit

* **New Features**
  * New accounts can now be signed in immediately after completing registration.
  * Successful sessions are automatically started and the user is marked online.
  * Login activity and call counts are recorded consistently for new and returning users.

* **Bug Fixes**
  * Registration and authentication flows now correctly preserve the newly created account through the login process.

* **Tests**
  * Added coverage for session initialization, login history updates, account consistency, and unknown accounts.

### DIFF
--- a/cmd/vision3/main.go
+++ b/cmd/vision3/main.go
@@ -1072,7 +1072,7 @@ func sessionHandler(s ssh.Session) {
 			}
 		}
 
-		matrixAction, matrixErr := menuExecutor.RunMatrixScreen(s, terminal, userMgr, int(nodeID), effectiveMode, int(termWidth.Load()), int(termHeight.Load()))
+		matrixAction, matrixUser, matrixErr := menuExecutor.RunMatrixScreen(s, terminal, userMgr, int(nodeID), effectiveMode, int(termWidth.Load()), int(termHeight.Load()))
 		if matrixErr != nil {
 			slog.Error("matrix screen error", "node", nodeID, "error", matrixErr)
 			return
@@ -1080,6 +1080,18 @@ func sessionHandler(s ssh.Session) {
 		if matrixAction == "DISCONNECT" {
 			slog.Info("user selected disconnect from matrix", "node", nodeID)
 			return
+		}
+		// A completed signup hands back the account to continue as, so the
+		// caller is not asked for credentials they set moments ago. The login
+		// loop below is skipped, but everything after it — including the login
+		// sequence — runs exactly as for any other caller.
+		if matrixUser != nil {
+			authenticatedUser = matrixUser
+			bbsSession.Mutex.Lock()
+			bbsSession.User = authenticatedUser
+			bbsSession.Mutex.Unlock()
+			userMgr.MarkUserOnline(authenticatedUser.ID)
+			slog.Info("continuing as newly created user", "node", nodeID, "user", authenticatedUser.Handle)
 		}
 		// matrixAction == "LOGIN" — proceed to normal login loop
 	}
--- a/internal/menu/executor_login_auth.go
+++ b/internal/menu/executor_login_auth.go
@@ -93,7 +93,7 @@ func runAuthenticate(c *cmdCtx, args string) (*user.User, string, error) {
 		}
 		// Carry them into the session rather than making them re-enter the
 		// handle and password they set moments ago.
-		if started := e.continueAsNewUser(userManager, newUser, nodeNumber); started != nil {
+		if started := e.continueAsNewUser(s, userManager, newUser, nodeNumber); started != nil {
 			return started, "", nil
 		}
 		return nil, "", nil // Could not continue: back to the LOGIN screen
--- a/internal/menu/executor_login_auth.go
+++ b/internal/menu/executor_login_auth.go
@@ -84,14 +84,19 @@ func runAuthenticate(c *cmdCtx, args string) (*user.User, string, error) {
 	// Check if user wants to apply as a new user
 	if strings.EqualFold(username, "new") {
 		slog.Info("user typed 'new' in AUTHENTICATE - starting new user application", "node", nodeNumber)
-		newUserErr := e.handleNewUserApplication(s, terminal, userManager, nodeNumber, outputMode, termWidth, termHeight)
+		newUser, newUserErr := e.handleNewUserApplication(s, terminal, userManager, nodeNumber, outputMode, termWidth, termHeight)
 		if newUserErr != nil {
 			if errors.Is(newUserErr, io.EOF) {
 				return nil, "LOGOFF", io.EOF
 			}
 			slog.Error("new user application error", "node", nodeNumber, "error", newUserErr)
 		}
-		return nil, "", nil // Return to LOGIN screen after signup
+		// Carry them into the session rather than making them re-enter the
+		// handle and password they set moments ago.
+		if started := e.continueAsNewUser(userManager, newUser, nodeNumber); started != nil {
+			return started, "", nil
+		}
+		return nil, "", nil // Could not continue: back to the LOGIN screen
 	}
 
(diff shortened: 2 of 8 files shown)


robbiew:
All four addressed in `890219b`. One was a security regression this PR introduced.

## IP lockout bypass (CodeRabbit, Major) — confirmed, and mine

The lockout check in both login paths sits **after** the `"new"` branch:

```text
executor_login_auth.go    line  85: if username == "new"  → signup
                          line 135: IsIPLockedOut
executor_login_prompt.go  line  79: if username == "new"  → signup
                          line 127: IsIPLockedOut
```

That ordering was harmless while signup could only return to the login screen — it authenticated nobody. This PR made signup hand back a session, so a locked-out client could register and walk straight past the lockout.

The check now lives inside `continueAsNewUser`, so all three entry points are covered rather than each having to remember, and a blocked continuation is logged via `logging.Security`.

## Deleted accounts in `BeginSession` (Copilot) — agreed

`Authenticate` checks `DeletedUser` before verifying the password, so nothing reachable today could exploit it. But extracting the stamping moved the guard **off the shared path**, which is exactly how a future caller would quietly acquire the hole. The check belongs where the mutation happens, and it is there now, with a test asserting a rejected session stamps no bookkeeping.

## Discarded pause error (CodeRabbit, Minor) — agreed, and worth spelling out

The account is created and saved *before* the pause, so a caller who dropped there still had a session started. The account existing is fine; the bookkeeping is not:

```text
phantom call recorded  → lastLogin = drop time, previousLogin = zero
their next real login  → previousLogin = the phantom time
```

On their first genuine call, "new since your last visit" would be measured from the moment they disconnected during signup, so every news item predating their registration reads as already seen — the `PRINTNEWS` failure from #198 arriving by another route. The error is returned now, and all three callers already stop on `io.EOF` before reaching continuation.

## Misleading test (Copilot) — agreed

`TestAuthenticateAndBeginSessionAgree` claimed to pin the two paths together without ever calling `Authenticate`. It drives both now and compares what each recorded, which is what the extraction was for in the first place.

Re-verified live afterwards: signup still continues into the session, with `timesCalled=1` recorded.

robbiew:
All four review findings are addressed in `890219b`:

- **`BeginSession` and soft-deleted users** (Copilot) — the `DeletedUser` guard now sits in `BeginSession` itself, before any mutation. You were right that leaving it only in `Authenticate` made the shared entry point a bypass waiting for its second caller. `TestBeginSessionDeniesDeletedAccounts` pins it, and asserts `TimesCalled` is untouched on rejection.
- **Misleading test name** (Copilot) — `TestAuthenticateAndBeginSessionAgree` now actually calls `Authenticate` and compares both paths, so the name and the body match.
- **IP lockout before `BeginSession`** (CodeRabbit, Major) — genuine security regression, and the sharpest catch here. The lockout check sits *after* the `"new"` branch in both login paths; that was harmless only while signup authenticated nobody, and this PR made it reachable. `continueAsNewUser` now checks `IsIPLockedOut(remoteIPFromSession(s))` and refuses the continuation before any session bookkeeping.
- **Final signup-pause error** (CodeRabbit) — `io.EOF` is returned so the callers' existing disconnect handling runs before continuation. Non-EOF read errors are logged and still continue, since the account is created and saved either way and a transient read hiccup should not strand a real user at the door.

Neither bot re-reviewed the follow-up commit, so I checked each finding against the code directly rather than waiting on a second pass.

PR #230, opened 2026-09-07
fix(menu): expand placeholders inside %%file.ans%% includes

### LINKED ISSUE #211: Placeholders don't expand inside %%file.ans%% includes; the markup renders as literal text
Found while implementing optional groups (#210). Filing separately because the scope turned out to be much wider than that feature.

## Problem

`displayPrompt` processes a menu prompt in this order (`internal/menu/executor_display.go`):

1. `expandOptionalGroups` — optional `|{...|}` groups
2. `|XX` placeholder substitution (`|UH`, `|UN`, `|GL`, `|TL`, `|LCALL`, …)
3. `@CODE@` AT-codes (`@UC@`, `@U@`)
4. **`processFileIncludes`** — `%%file.ans%%` tags replaced with file content
5. `@RR@` random rumor
6. `ansi.ReplacePipeCodes`

Because includes are expanded at step 4, everything resolved in steps 1–3 has already run. Any placeholder living inside an included file is never substituted.

It does not fail loudly either. Unrecognized `|XX` sequences pass straight through `ReplacePipeCodes` as literal text:

```
"|UH"        -> "|UH"
"|UN"        -> "|UN"
"|LCALL"     -> "|LCALL"
"|{(|UN)|}"  -> "|{(|UN)|}"
```

So a sysop who puts `|UH` in an included ANSI file sees the literal characters `|UH` on screen instead of the caller's handle, with nothing in the log to explain it.

## This is an inconsistency, not a deliberate design

`@RR@` is expanded *after* includes, with an explicit comment saying why:

```go
// 2b. Expand @RR@ after file includes so %%file.ans%% content is also processed.
```

So the need for post-include processing was already recognised for one code. Every other placeholder was left on the wrong side of the boundary.

## Severity

Low today: no menu in the shipped `menus/v3` set uses `%%file.ans%%` includes at all, so nothing in the default configuration is affected. This bites a sysop who reaches for the include feature and reasonably expects placeholders to work there.

## Options

- **Move `processFileIncludes` before substitution.** Simplest, and makes every placeholder work uniformly. Needs a check that the recursion guard still behaves, since included content would then be scanned for placeholders.
- **Re-run substitution after includes.** More surgical, but placeholder *values* would then be present in the text before the include pass — worth confirming a value cannot itself contain a `%%…%%` tag and trigger an unintended include. `|GL` and `|UN` are user- and sysop-settable, so this needs care.

Either way the ordering should be stated in one place rather than implied by the sequence of calls.

## Also

Document the behaviour in the menu system reference next to the placeholder tables, whichever way it is resolved — right now nothing tells an author that includes and placeholders do not compose.

Related: #209 (optional groups in message header templates), #210 (optional groups in prompts, where this was found).

robbiew:
Closes #211.

Includes were resolved **after** substitution, so nothing inside an included file was ever expanded. And because an unrecognised `|XX` code passes through `ReplacePipeCodes` untouched, the markup appeared verbatim on screen with nothing in the log to explain it.

## Before and after, same include

```text
before  [inc] handle=|UH level=|LEVEL loc=|GL users=@UC@|{ note=(|UN)|}
after   [inc] handle=Felonius level=255 loc=ViSiON/3 users=1
```

Captured live by pointing `MAIN.MNU`'s prompt at an include and running the pre-fix binary against the same file. Every form of markup was leaking to the screen — `|XX` placeholders, the `@UC@` AT-code, and `|{...|}` optional groups.

## The fix, and why front rather than back

Includes now run at the front of the pipeline:

```text
includes → optional groups → |XX substitution → @CODE@ → @RR@ → pipe codes
```

so included content is treated exactly like text written inline in the prompt.

Re-running substitution *after* includes would also have worked, but front is safer. With includes resolved before any value is substituted, a placeholder whose **value** contains `%%something.ans%%` cannot pull in a file. `|GL` and `|UN` are user-settable, so that ordering is not hypothetical — and this removes the possibility rather than relying on values never happening to look like an include tag.

`@RR@` no longer needs its position at the end. It was expanded after includes specifically so included content was covered, which the new order gives it anyway; the comment explaining that is gone with the reason for it.

## Docs

The menu reference stated that groups did not work inside includes, and that no placeholder did. Both are now untrue, so the note is replaced with what the behaviour actually is.

## Testing

`go build ./...` and `go test ./...` clean. Verified live in both directions with an include exercising a placeholder, an AT-code and an optional group — the before/after above is the real render from the same file against the two binaries.

## Summary by CodeRabbit

* **Bug Fixes**
  * Included prompt files now correctly support optional groups, placeholders, pipe codes, and AT-codes.
  * Prompt content remains intact when an included file is missing.
  * Prevented placeholder values from being misinterpreted as file includes.
  * Improved handling of nested file inclusions and overlapping placeholders.
  * Self-referencing included files now terminate safely.

* **Documentation**
  * Updated optional-group documentation to reflect full markup support within included files.

### DIFF
--- a/docs/sysop/menus/menu-system.md
+++ b/docs/sysop/menus/menu-system.md
@@ -362,11 +362,9 @@ Rules:
 - Groups do not nest. An unmatched `|{` is left as visible text rather than
   swallowing the rest of the prompt.
 
-Groups are resolved on the prompt string itself, before `%%file.ans%%` includes
-are expanded — so `|{...|}` works in `PROMPT1`/`PROMPT2` but not inside an
-included ANSI file. This is not specific to groups: **no** `|XX` placeholder
-expands inside an included file, and the markup renders as literal text rather
-than failing loudly. See issue #211.
+`%%file.ans%%` includes are resolved first, so an included file behaves exactly
+like text written inline in the prompt: groups, `|XX` placeholders and `@CODE@`
+AT-codes all work inside one.
 
 The `|{P}` and `|{O}` login position markers share the `|{` prefix but are not
 groups; they are stepped over and pass through untouched.
--- a/internal/menu/executor_display.go
+++ b/internal/menu/executor_display.go
@@ -499,13 +499,11 @@ const maxIncludeRounds = 6
 // It now looks for included files within the MENU SET's ansi directory.
 func (e *MenuExecutor) processFileIncludes(prompt string, depth int) string {
 	if depth >= maxIncludeRounds {
-		slog.Warn("exceeded maximum file inclusion depth, stopping processing", "maxRounds", maxIncludeRounds)
+		slog.Warn("exceeded maximum file inclusion rounds, stopping processing", "maxRounds", maxIncludeRounds)
 		return prompt
 	}
 
-	processedAny := false
 	result := includeTagRe.ReplaceAllStringFunc(prompt, func(match string) string {
-		processedAny = true
 		// match is "%%name.ext%%"; strip the delimiters instead of re-matching.
 		fileName := strings.TrimSuffix(strings.TrimPrefix(match, "%%"), "%%")
 		// Look for included file in MenuSetPath/ansi
@@ -520,7 +518,12 @@ func (e *MenuExecutor) processFileIncludes(prompt string, depth int) string {
 		return string(data)
 	})
 
-	if processedAny {
+	// Recurse on what is left to do, not on what was just done. Recursing
+	// because this round expanded something costs a wasted scan on every
+	// prompt with includes, and worse, a legitimate nest exactly at the cap
+	// would expand its last tag and then recurse once more purely to trip the
+	// limit -- logging "exceeded maximum" about a file set that was fine.
+	if includeTagRe.MatchString(result) {
 		return e.processFileIncludes(result, depth+1)
 	}
 
--- a/internal/menu/executor_display.go
+++ b/internal/menu/executor_display.go
@@ -402,57 +402,12 @@ func (e *MenuExecutor) displayPrompt(terminal *term.Terminal, menu *MenuRecord, 
 		}
 	} // End if currentUser != nil
 
-	// Pull in %%file.ans%% content first, so everything below treats it exactly
-	// like text written inline in the prompt. Includes used to run after
-	// substitution, which meant no |XX placeholder inside an included file was
-	// ever expanded — and since an unrecognised code passes through
-	// ReplacePipeCodes untouched, the markup appeared verbatim on screen with
-	// nothing in the log to explain it.
-	//
-	// Front rather than back for safety as well as simplicity: with includes
-	// resolved before substitution, a placeholder whose *value* happens to
-	// contain "%%something.ans%%" cannot pull in a file. |GL and |UN are
-	// user-settable, so that ordering matters.
-	promptString, err := e.processFileIncludes(promptString, 0)
-	if err != nil {
-		slog.Error("failed processing file includes in prompt", "menu", currentMenuName, "error", err)
-		return err
-	}
-
-	// Drop |{...|} groups whose placeholders are all empty, so decoration around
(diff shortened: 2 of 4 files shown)


robbiew:
Both addressed in 1a96b89.

**Regression test** — you were right that nothing pinned this. The ordering is load-bearing and fails quietly when wrong, which is exactly how it survived unnoticed.

I lifted the pipeline out of `displayPrompt` into `renderPromptText` so it is testable without a terminal, and added `internal/menu/prompt_pipeline_test.go` with five cases: included content fully expanded (`|XX`, `@UC@` and `|{...|}` all inside the include), an optional group inside an include kept when populated, longest-first substitution, a missing include leaving the rest of the prompt intact, and one asserting a placeholder value that *looks* like an include tag cannot pull in a file — `|GL` and `|UN` are user-settable, so include-before-substitute is what keeps that unreachable.

I checked the test actually catches the regression rather than just passing: restoring the old order temporarily fails it with the same symptom seen live —

```
"|UH" reached the output unexpanded: "[handle=|UH level=|LEVEL users=@UC@|{ note=(|UN)|}]"
```

**Dead error path** — confirmed, every return in `processFileIncludes` was `nil`. I dropped the return rather than propagating real errors: a missing include is logged and skipped by design, and a typo in an include tag should not blank an entire prompt. Signature is now `processFileIncludes(prompt string, depth int) string`, with the two test call sites updated.

robbiew:
Both addressed in `b951d2d`.

**Shared ordering** — agreed, and the helper's own doc says it "mirrors how substitution itself resolves prefix collisions", so the duplication was exactly the drift risk you describe. `renderPromptText` now calls `placeholderKeysLongestFirst` instead of carrying its own sort. `TestLongerPlaceholdersSubstituteFirst` covers the behaviour either way.

**Depth cap** — you're right that it read as one lower than it allowed: `depth` starts at zero and the test was `>`, so a constant of 5 permitted six rounds. I kept the count and fixed the naming rather than tightening the limit — the guard exists to stop a cycle, not to be a meaningful nesting budget, and quietly reducing an existing limit could break a menu set that happens to sit at the boundary. It is now `maxIncludeRounds = 6` with an exclusive `>=` test, named for what it actually counts.

While there I noticed nothing covered the case the cap exists for, so I added `TestSelfIncludingFileTerminates` — a file that includes itself, asserting it terminates and within a bounded number of rounds.

robbiew:
Addressed in `67ebf98`.

You're right that the threshold was permissive enough to be nearly decorative — six rounds are allowed and the test tolerated sixteen, so a regression expanding seven through sixteen times would have sailed through.

`maxIncludeRounds` moves to package scope and the test now asserts against it exactly, so the assertion tracks the configuration rather than a number that happens to sit above it. I also added an assertion that the tag which hit the cap is left in place rather than dropped, so a menu set that trips the limit shows evidence of it instead of silently losing content.

Checked it actually bites by restoring the off-by-one this PR fixes:

```
--- FAIL: TestSelfIncludingFileTerminates
    expanded 7 rounds, want exactly 6: "xxxxxxx%%loop.ans%%"
```

robbiew:
Both addressed in `072782e`. The first was a real behavioural bug, not just wasted work.

**Recursing on work done rather than work remaining.** You were right that this can trip the cap in non-cyclic cases, and I confirmed it before fixing: a legitimate chain nested exactly six deep expands its last tag, recurses once more with nothing left to do, and logs

```
level=WARN msg="exceeded maximum file inclusion rounds, stopping processing" maxRounds=6
```

about a menu set that was entirely fine. A sysop reading that would go looking for a cycle that does not exist. It now recurses only while the result still holds a tag, which also removes the wasted scan on every prompt with an include.

`TestNestingExactlyAtTheCapExpandsWithoutWarning` covers it, and fails against the previous version with exactly that warning.

**Warning wording** — fixed, it said "depth" after the constant became a round count.

robbiew:
@coderabbitai review

robbiew:
@coderabbitai full review

PR #232, opened 2026-09-07
fix(users): lock users.json across processes and write it atomically

### LINKED ISSUE #216: users.json has no cross-process lock: TOCTOU between the BBS and ./ue
Raised in review of PR #214 and deliberately left out of it.

## Problem

PR #214 makes the BBS detect that `./ue` has edited `users.json` and merge those edits instead of overwriting them. The sequence on save is:

1. fingerprint the file, see it changed
2. read it and merge the sysop's fields in
3. write the whole map back

Nothing stops `./ue` writing between steps 1 and 3. If it does, the BBS overwrites the newer editor data, which is the same class of loss #214 set out to fix — just through a much narrower window.

The window is small and both writers are infrequent, so this is far less likely than the bug #214 fixes (which was deterministic, not a race). But it is real.

## Why it was not fixed in #214

Closing it properly means an advisory lock held across read-merge-write and honoured by both processes. That needs:

- a locking scheme both the BBS and `./ue` use — `flock` on unix, `LockFileEx` on Windows, or a portable lock-file protocol
- `./ue` taught to take the same lock (it currently uses its own mtime check, `usereditor/fileio.go:44`)
- a decision on behaviour when the lock is held: block, retry, or refuse with a message

That is a cross-cutting change to two processes and both platforms, which does not belong inside a fix for the deterministic bug.

## Related

`internal/user/manager_persist.go` also notes that `os.WriteFile` truncates in place and is not atomic — a crash mid-write can leave a truncated `users.json`. A write-to-temp-then-rename would fix that, and is naturally done alongside locking since both concern how the file is written. `./ue` already writes atomically (`usereditor/fileio.go:52`), so again the editor is ahead of the BBS here.

Worth doing together.

robbiew:
Closes #216.

## The race

The BBS folds in `./ue`'s edits before rewriting `users.json`, but nothing holds the file between the check and the write:

1. fingerprint the file, see it changed
2. read it and merge the sysop's fields in
3. write the whole map back

An editor write landing between 1 and 3 is destroyed — the same loss the merge exists to prevent, through a narrower window.

## What this does

**`internal/filelock`** — the advisory lock both processes take. `flock` on unix, `LockFileEx` on Windows, lifted from the door-lock helpers that were already a duplicated pair.

The lock sits on a `users.json.lock` **sidecar**, not on `users.json`. Writers replace the file by rename, which swaps the inode, so a lock on the data file would leave two processes holding locks on different objects the moment either saved. The sidecar is never renamed — and never deleted, since unlinking after unlock races with the next process to take the same path.

**BBS** — holds the lock across the whole read-merge-write. If it cannot be taken in five seconds it saves anyway and logs: that is the pre-lock behaviour, and a small race beats dropping a session's data.

**Atomic writes** — `os.WriteFile` truncates then writes, so a crash between the two leaves a truncated `users.json`. Saves now go temp-file → sync → rename. `./ue` has always written this way.

## Two `./ue` defects this surfaced

**Its changed-file check used mtime**, which cannot see a BBS write in the same filesystem tick as the load. Both processes now share one content-based definition via `user.FingerprintBytes`.

**Confirming "File modified externally! Overwrite?" wrote nothing.** The handler called the same *guarded* save, which re-ran the check that raised the prompt, failed it again and returned — sending the sysop back to the list with their edits silently dropped and no second warning. Check and write now happen together under one lock, with an explicit `force`.

## Two consequences worth knowing

- `users.json` is now `0600` on **every** save rather than only at creation. That is what the code always asked for and is right for a file of password hashes, but it tightens permissions on existing installs.
- A read-only `users.json` no longer makes a save fail — rename needs write permission on the *directory*. `breakUserStore` in the menu tests moves accordingly, to where an atomic write actually fails.

## Verification

Live against the local BBS, with `./ue`'s protocol driven from a **separate python process using `flock(2)` directly**, so what was tested is the on-disk convention rather than this package talking to itself:

- lock held past the timeout → the BBS logged the documented fallback twice, five seconds apart, proving it really waits
- lock held a realistic 200ms → login clean, no warning, external demotion survived
- `./ue` loads and renders the merged record

Every test was checked against the unfixed code rather than trusted for passing:

```
--- FAIL: TestSaveIsNeverObservedPartiallyWritten
    reader saw a partially written file: unexpected end of JSON input
--- FAIL: TestConfirmingTheOverwritePromptWrites
    confirming the overwrite wrote nothing: level on disk = 99, want 42
```

Cross-process exclusion is tested by re-executing the test binary as a child that holds the lock — an in-process test could not prove it.

## Summary by CodeRabbit

- **New Features**
  - Detects external changes to user data before saving, even when changes occur rapidly.
  - Prompts for confirmation before overwriting externally modified data, with an option to force the save.
  - Coordinates saves across processes to reduce conflicts between simultaneous edits.

- **Bug Fixes**
  - User data is now saved atomically, preventing empty or partially written files.
  - Successful saves refresh file-change tracking and preserve edits made during concurrent operations.

### DIFF
--- a/internal/filelock/filelock.go
+++ b/internal/filelock/filelock.go
@@ -1,0 +1,93 @@
+// Package filelock provides an advisory cross-process lock, used to serialise
+// read-modify-write cycles on a shared data file between the BBS and the
+// separate editor binaries.
+//
+// The lock is advisory: it only excludes processes that ask for it. It is not
+// a permission check and does not stop an unrelated program from writing the
+// file underneath us.
+//
+// internal/menu keeps its own lock helpers for single-instance doors. That is
+// a different job -- a reservation held for as long as a door runs, carrying
+// node and pid metadata inside the lock file for diagnosing stale ones -- and
+// folding it in here would mean exposing the file handle just for that. These
+// two are deliberately separate rather than an oversight.
+package filelock
+
+import (
+	"fmt"
+	"os"
+	"time"
+)
+
+// DefaultTimeout is a sensible wait for a data-file lock. Holders take the
+// lock across a read-merge-write of a small JSON file, which is milliseconds,
+// so anything approaching this means a wedged or stopped process rather than
+// ordinary contention.
+const DefaultTimeout = 5 * time.Second
+
+// retryInterval is how often a blocked acquire re-tries. The underlying
+// primitives are non-blocking so that Acquire can honour a timeout at all;
+// polling is what buys that. Short enough to be imperceptible, long enough not
+// to spin a core while waiting.
+const retryInterval = 20 * time.Millisecond
+
+// Lock is a held advisory lock. Release it when the critical section ends.
+type Lock struct {
+	f *os.File
+}
+
+// SidecarPath returns the lock file path guarding dataPath.
+//
+// The lock deliberately lives on a sidecar rather than on the data file
+// itself. Writers here update the data file atomically, by writing a temp file
+// and renaming it into place, which replaces the inode. A lock taken on the
+// data file is attached to the open file description behind the *old* inode,
+// so the instant a writer renamed over it the two processes would be holding
+// locks on different objects and both would proceed. The sidecar is never
+// renamed, so every participant contends on the same inode.
+func SidecarPath(dataPath string) string {
+	return dataPath + ".lock"
+}
+
+// Acquire takes an exclusive lock guarding dataPath, waiting up to timeout.
+// A timeout of zero or less tries exactly once.
+//
+// The returned error distinguishes nothing about why the wait failed on
+// purpose: to a caller, "someone else holds it" and "we waited long enough"
+// are the same decision.
+func Acquire(dataPath string, timeout time.Duration) (*Lock, error) {
+	lockPath := SidecarPath(dataPath)
+	f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0600)
+	if err != nil {
+		return nil, fmt.Errorf("open lock file %s: %w", lockPath, err)
+	}
+
+	deadline := time.Now().Add(timeout)
+	for {
+		if tryLock(f) {
+			return &Lock{f: f}, nil
+		}
+		if !time.Now().Before(deadline) {
+			_ = f.Close() // not holding the lock; nothing to release
+			return nil, fmt.Errorf("timed out after %s waiting for lock %s", timeout, lockPath)
+		}
+		time.Sleep(retryInterval)
+	}
+}
+
+// Release drops the lock. Safe to call on a nil Lock so callers can defer it
+// next to an acquire whose error they handle by carrying on unlocked.
+//
+// The lock file is deliberately left on disk. Removing it after unlocking
+// opens a race where another process locks the same path between our unlock
+// and our remove, and we then unlink the file they are holding — after which a
+// third process creates a fresh one and two writers believe they hold the same
+// lock. An empty sidecar file costs nothing.
+func (l *Lock) Release() {
+	if l == nil || l.f == nil {
+		return
+	}
+	unlock(l.f)
+	_ = l.f.Close() // lock already dropped above
+	l.f = nil
+}
--- a/internal/filelock/filelock_test.go
+++ b/internal/filelock/filelock_test.go
@@ -1,0 +1,139 @@
+package filelock
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+func TestAcquireAndRelease(t *testing.T) {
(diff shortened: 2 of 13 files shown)


robbiew:
All addressed in `4178824`, except the Windows rename point — reasoning below.

**Quitting over an unanswered prompt** (both of you, and the sharpest catch here). This is the same defect as the forced-overwrite bug this PR already fixes, one layer out, and I missed that it generalised. `modeExitConfirm` and `modeSaveConfirm` returned `tea.Quit` unconditionally, so a save that raised the "modified externally" prompt was quit straight over the top of it — unanswerable, edits gone. A plain `SAVE ERROR` was discarded the same way, unread.

Quitting is now gated on the save landing, and the exit intent is remembered, so answering the prompt *finishes the exit* rather than stranding the sysop back in the list — which is the behaviour @coderabbitai specifically asked for. Any other failure returns to the list still dirty with the error showing, so the work is recoverable. `TestExitConfirmStaysOpenOnAConflictThenQuits` pins it; against the unfixed code it reports `the editor quit over the top of the overwrite prompt, losing the edits`.

**`0600` on the editor's temp file** — correct, and an inconsistency with my own BBS-side writer, which already chmods. `os.CreateTemp` asks for 0600 but the umask still applies and rename carries the result onto `users.json`, after which both processes can fail to read it. Fixed.

**Stale `reloadPersistedUser` doc** — right, it restores the directory now, not the file. (The `breakUserStore` block above it already described the new behaviour correctly; that is the one Copilot pinned, but the stale text was in the function below.)

**`breakUserStore` on Windows** — good catch that it cannot work there: the directory stays writable, the save succeeds, and the rollback path silently goes untested. It now skips on Windows, which is honest about not covering it. CI is `ubuntu-latest` only, so this changes nothing there.

**`os.Rename` atomicity on Windows** — not doing this one, and I want to be explicit rather than quietly skip it. Go's `os.Rename` on Windows is `MoveFileEx(..., MOVEFILE_REPLACE_EXISTING)`, which is the standard atomic-replace approach on a single volume and what Go's own ecosystem uses; `ReplaceFile` buys attribute/ACL preservation rather than a stronger atomicity guarantee for this case. Adding a platform-specific path here would be real complexity on a platform CI never exercises, to harden something already at parity with normal practice.

The genuine Windows wrinkle is different from the one described: `os.Rename` fails outright if another process holds the destination open without `FILE_SHARE_DELETE`. That is availability, not a torn write, and neither process holds `users.json` open across a save. Happy to file a follow-up issue for Windows write-path hardening if you would like it tracked — say the word and I will.

robbiew:
Addressed in `11436c2`.

Good catch, and the distinction is the point: `CheckFileChanged` returned "unchanged" for anything it could not fingerprint, which lumped *a file that is not there* together with *a file that exists and cannot be read*. Those want opposite answers.

A missing file genuinely has nothing to overwrite. A file we could not open has unknown contents that the save is about to replace, and "unknown" is not a reason to assume "ours" — it skipped the overwrite prompt and destroyed whatever was there. A permission or I/O error now raises the prompt and lets the sysop decide; missing still reports unchanged, so the no-way-out case my comment worried about stays covered.

Two tests, both skipping as root and on Windows where the permission bits they rely on do not hold. Against the previous behaviour they report:

```
an unreadable users.json reported unchanged; the next save would overwrite it blind
guarded save error = <nil>, want ErrFileChanged
```

robbiew:
@coderabbitai review

robbiew:
Filed the Windows follow-up as #235, as offered.

It turned out to deserve more than the `ReplaceFile` question on its own: Windows is a **released platform** (`README.md:179`), but CI is `ubuntu-latest` only, so five Windows-tagged source files ship without CI ever executing them — including `internal/filelock/filelock_windows.go` from this PR.

That is the part I would fix first, ahead of the rename question. This PR introduced the lock to stop the BBS and `./ue` destroying each other's writes, and the unix path is covered down to a child-process test proving real cross-process exclusion. The `LockFileEx` path has never run. A lock that silently fails to exclude is worse than no lock, because the design now assumes it holds.

On the original point, the issue records where I agree and where I do not: `os.Rename` on Windows is `MoveFileEx(..., MOVEFILE_REPLACE_EXISTING)`, which is the standard atomic-replace approach, and `ReplaceFile` mainly buys ACL/attribute/stream preservation. The genuine Windows failure mode looks like availability rather than corruption — rename fails outright if another process holds the destination open without `FILE_SHARE_DELETE`, which on a multi-node BBS plus `./ue` is plausible. Also noted that this predates the PR: `internal/menu/infoforms_store.go:93` already carried the caveat, across 39 rename sites.

Suggested order there is CI matrix first, then verify the lock, then decide on `ReplaceFile` informed by both rather than in the abstract.

PR #233, opened 2026-09-07
fix(users): refresh the session's user record from disk at menu entry

### LINKED ISSUE #215: Live session keeps a stale user copy after an external edit
Split out of #199 / PR #214, which fixes persistence. This is the remaining behavioural gap.

## Problem

With #214, a sysop editing a user in `./ue` while the BBS is running now has that edit persist correctly — it is no longer reverted by the next save.

But the user's *running session* still holds the copy it took at authentication. So:

- demote someone to level 20 while they are online, and they keep level 255 access for the rest of that call
- validate someone mid-session, and they still cannot reach validated-only areas until they log off and back on

The change is correct on disk and takes effect at their next login. It just does not reach the session already in progress.

## Why it is separate

#214 is about not losing data. This is about propagating a change into live state, which needs different machinery: a session's `currentUser` is passed by value through the menu executor and the login sequence, so there is no single place to refresh it.

## Options

- **Refresh at menu transitions.** Re-read the user from `UserMgr` when entering a menu, so a change lands within a keystroke or two. Cheap, and menu entry is already a natural checkpoint. Needs care that it does not clobber unsaved session state.
- **Notify the session.** Have the merge in #214 signal affected sessions so they refresh on the spot.
- **Leave it, and tell the sysop.** Document that edits apply at next login, and surface it in `./ue` once online state is visible to it (see below).

## Related

`./ue` cannot currently tell whether a user is online at all: `activeUserIDs` lives in memory inside the BBS process (`internal/user/manager_stats.go:39`) and is never written to disk. If the BBS published node/session state, `./ue` could warn "this user is online; changes apply at their next login" — which is useful regardless of which option above is chosen.

robbiew:
Closes #215. **Stacked on #232** (base is that branch) — the disk sync uses `internal/filelock` from it. Rebases onto `main` once #232 lands.

## The problem

A session's user record is a copy taken once at authentication and carried by value through the login sequence and the menu executor. It never changes again, so a sysop editing an account in `./ue` while its owner is online has no effect on them until they hang up and dial back.

It is **not** an access-control problem, which is how it tends to get described. The whole record is frozen — setting a message header style or a password mid-session is ignored exactly as thoroughly as a demotion, and nothing indicates the value on disk is being ignored.

Adding `users.json` to the config watcher looks like the fix and is not: that refreshes the manager's *map*, and the session does not read from the map. Two separate staleness problems; this is the second.

## The fix

`UserMgr.RefreshSessionUser` returns an updated copy, taking exactly `sysopOwnedFields` from the stored record and leaving everything else alone. The rest is session-owned and the copy in hand is the newer authority for it — times called, downloads, file points, newscan pointers, tagged files have all moved on since authentication, and pulling them from disk would discard the current call.

**Menu entry is the checkpoint.** There is no single place to push a change into a value-copied record, so re-reading where the loop already restarts means an edit lands within a keystroke or two. A demotion also re-arms the idle timeout, which derives from access level — without that, someone demoted out of sysop keeps an unlimited session for the rest of the call.

The poll is throttled to once a second across the process: refreshing at every menu puts it on the path of each keystroke that changes screen, and the check hashes the file.

## A second bug found on the way

`./ue` soft-deletes by setting `DeletedUser` on the record rather than removing it, and `sysopOwnedFields` **excluded deletion state** — so the next save from a running BBS wrote the account back undeleted. A sysop removing an abusive user mid-call had it quietly undone:

```
DeletedUser on disk after a BBS save = false (ue set it true)
  -> the BBS reverted the editor's soft delete
```

Deletion state now merges from disk. Without it the new deleted-account check could not fire for the way sysops actually delete. Flagging it explicitly since it widens `sysopOwnedFields` — happy to split it out if you would rather.

## Verified live

Demoting on disk while sitting at the main menu:

```
access level changed underneath a live session handle=Felonius node=1 from=255 to=20
```

Soft-deleting mid-call:

```
account removed or deleted while online; ending the call handle=Felonius node=1
```

An ordinary login is unaffected — authenticated once, no calls ended, main menu reached.

## Still open from the issue

`./ue` cannot tell whether a user is online at all (`activeUserIDs` lives in memory and is never written to disk), so it cannot warn "this user is online". That is listed under *Related* in the issue and is untouched here.

## Summary by CodeRabbit

- **New Features**
  - Active sessions now automatically reflect updates made to user profiles and access levels.
  - Account renames are recognized without requiring users to log in again.
  - Session settings remain intact while profile changes are synchronized.

- **Bug Fixes**
  - Deleted or removed accounts are now promptly signed out with an access-denied message.
  - Changes to permissions automatically reapply the appropriate idle-timeout rules.
  - Sessions log off when user information cannot be refreshed safely.

### DIFF
--- a/internal/filelock/filelock.go
+++ b/internal/filelock/filelock.go
@@ -1,0 +1,93 @@
+// Package filelock provides an advisory cross-process lock, used to serialise
+// read-modify-write cycles on a shared data file between the BBS and the
+// separate editor binaries.
+//
+// The lock is advisory: it only excludes processes that ask for it. It is not
+// a permission check and does not stop an unrelated program from writing the
+// file underneath us.
+//
+// internal/menu keeps its own lock helpers for single-instance doors. That is
+// a different job -- a reservation held for as long as a door runs, carrying
+// node and pid metadata inside the lock file for diagnosing stale ones -- and
+// folding it in here would mean exposing the file handle just for that. These
+// two are deliberately separate rather than an oversight.
+package filelock
+
+import (
+	"fmt"
+	"os"
+	"time"
+)
+
+// DefaultTimeout is a sensible wait for a data-file lock. Holders take the
+// lock across a read-merge-write of a small JSON file, which is milliseconds,
+// so anything approaching this means a wedged or stopped process rather than
+// ordinary contention.
+const DefaultTimeout = 5 * time.Second
+
+// retryInterval is how often a blocked acquire re-tries. The underlying
+// primitives are non-blocking so that Acquire can honour a timeout at all;
+// polling is what buys that. Short enough to be imperceptible, long enough not
+// to spin a core while waiting.
+const retryInterval = 20 * time.Millisecond
+
+// Lock is a held advisory lock. Release it when the critical section ends.
+type Lock struct {
+	f *os.File
+}
+
+// SidecarPath returns the lock file path guarding dataPath.
+//
+// The lock deliberately lives on a sidecar rather than on the data file
+// itself. Writers here update the data file atomically, by writing a temp file
+// and renaming it into place, which replaces the inode. A lock taken on the
+// data file is attached to the open file description behind the *old* inode,
+// so the instant a writer renamed over it the two processes would be holding
+// locks on different objects and both would proceed. The sidecar is never
+// renamed, so every participant contends on the same inode.
+func SidecarPath(dataPath string) string {
+	return dataPath + ".lock"
+}
+
+// Acquire takes an exclusive lock guarding dataPath, waiting up to timeout.
+// A timeout of zero or less tries exactly once.
+//
+// The returned error distinguishes nothing about why the wait failed on
+// purpose: to a caller, "someone else holds it" and "we waited long enough"
+// are the same decision.
+func Acquire(dataPath string, timeout time.Duration) (*Lock, error) {
+	lockPath := SidecarPath(dataPath)
+	f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0600)
+	if err != nil {
+		return nil, fmt.Errorf("open lock file %s: %w", lockPath, err)
+	}
+
+	deadline := time.Now().Add(timeout)
+	for {
+		if tryLock(f) {
+			return &Lock{f: f}, nil
+		}
+		if !time.Now().Before(deadline) {
+			_ = f.Close() // not holding the lock; nothing to release
+			return nil, fmt.Errorf("timed out after %s waiting for lock %s", timeout, lockPath)
+		}
+		time.Sleep(retryInterval)
+	}
+}
+
+// Release drops the lock. Safe to call on a nil Lock so callers can defer it
+// next to an acquire whose error they handle by carrying on unlocked.
+//
+// The lock file is deliberately left on disk. Removing it after unlocking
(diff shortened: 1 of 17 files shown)


robbiew:
Both addressed in `4da924b`. The first one was a genuine regression this branch introduced — thank you.

**Rename drops the call.** Confirmed before fixing, rather than reasoning about it:

```
after an external rename: stillValid=false
  -> the live session was terminated by a rename
```

The handle is the map key, so renaming in `./ue` re-keys the record — `mergeExternalEdits` adds the new key and drops the old one as an external delete. The handle-only lookup then missed and reported the account gone, so a sysop *correcting someone's handle* disconnected them. Worth being clear this is new damage from this branch: before the refresh existed, a rename simply had nothing to act on.

Lookup now falls back to the ID when the handle misses — stable across a rename, and how `UpdateUserByID` already locates records for the same reason — and the refreshed copy takes the stored handle so later saves reach the right record. A zero ID is refused rather than matched, since that is what an unsaved or synthetic record looks like.

Verified live: renaming *and* demoting in one edit while sitting at the main menu logs `from=255 to=20` with no dropped call.

**Disk I/O under `um.mu`.** Also correct, and worse than one read — it was two, since `fingerprintOf` re-read the file after the merge. On the path of every menu change on every node, that put each session's disk read in the way of all the others.

The file is now read and fingerprinted once outside the lock, parsed outside it, and only the merge runs under it. One thing I did differently from your suggestion: the fingerprint recorded is of the bytes actually merged, not a fresh read afterwards. A re-read could have moved on in between, which would leave the map and the recorded state describing different content and cause the next poll to re-merge. `readUsersFromDisk` splits into read and parse to make that possible.

Added `TestRefreshIsSafeUnderConcurrency` since this is now a hot multi-node path; passes under `-race`.

robbiew:
Both addressed in `456a497`.

**Blocking on the lock at menu transitions** — correct, and I had already watched this happen on the *save* path earlier in this stack: holding the lock past the timeout stalled the BBS for five seconds, twice. Waiting there is right, because the alternative is losing data. Waiting here is not: the refresh is a read, it already proceeds unlocked on failure, and there is nothing to gain by waiting since the poll comes round again a second later. Reads are safe unlocked regardless — writers replace the file by rename, so the worst an unlocked read sees is the version one save old.

It now makes a single attempt and moves on. `TestRefreshDoesNotWaitOnAHeldLock` pins it and reports `refresh blocked 5.006605791s` against the previous timeout. Re-verified live that an ordinary mid-session demotion still lands.

**Nil contract** — fixed. The doc now says false means "no record to refresh", covering both the login phase and a deleted account, and what each kind of caller should do with it.

robbiew:
@coderabbitai review

robbiew:
Added in `90fe78d`.

Fair point, and the gap was real: every test in that file calls `allowImmediateSync`, so `diskSyncInterval` was never exercised. Deleting the throttle outright would have left the suite green while `syncFromDisk` went back to reading and hashing `users.json` on every menu change on every node — exactly the cost it exists to avoid, and invisible in tests.

`TestRefreshThrottlesDiskReads` asserts an edit made inside the interval is not read, and does land once it expires. Verified against a build with the throttle removed:

```
--- FAIL: TestRefreshThrottlesDiskReads
    AccessLevel = 30, want 20 — the throttle did not suppress the second read
```

PR #236, opened 2026-09-07
fix(menu): print the running version, not one baked into strings.json

### LINKED ISSUE #229: Print proper version number in menu command "^" version
currently, it prints: ViSiON/3 Go Edition - v0.1.0 (Pre-Alpha). But that is not the actual version.

robbiew:
Closes #229.

## The bug

Reproduced live before touching anything — a 0.8.2 build advertising:

```
ViSiON/3 Go Edition - v0.1.0 (Pre-Alpha)
```

The version was written into the shipped `strings.json`, and nothing rewrites a config file on upgrade, so every install has carried that line since whenever it was true.

The runnable already knew how to fill a `%s` from `internal/version` (what release builds stamp). The shipped default simply had no `%s`, so the substitution never fired. The string editor's own metadata has described the key as `Version string format (%s=version)` the whole time — **the contract was right, the data didn't follow it**.

## The fix

The default now carries the verb. An install still holding a previously shipped banner is upgraded, since carrying one unchanged means no sysop ever edited it; anything else is left alone, because a customised banner is theirs. Superseded defaults are listed explicitly rather than matched loosely, so that stays true next time the default changes.

A template that can't be filled prints as-is with a warning naming the problem. `Sprintf` doesn't fail on wrong arity — it renders `%!s(MISSING)` into the middle of the sysop's banner, which is worse than leaving their text alone.

## A bug in my own guard

Writing that guard turned up a bug in it. Counting format verbs isn't enough: a banner ending `100% Go` has one verb by that measure, and it's `%<space>`. My first version still fed it to Sprintf:

```
"|15Trailing percent %!|(string=v0.8.2)07"
```

Caught by the test I'd just written. Only a template whose single verb is `%s` gets filled now.

## Verified live

All three states, against the local BBS:

| install state | `^` prints |
|---|---|
| still on the old baked-in banner | `ViSiON/3 Go Edition - v0.8.2` |
| customised (`The Dungeon :: %s`) | `The Dungeon :: v0.8.2` |
| no verb (`100% Go, no verb here`) | printed unchanged, warning logged |

The warning is specific enough to act on:

```
execVersionString must contain exactly one %s and nothing else; printing it
unchanged to avoid a malformed banner  verbs=1 stringVerbs=0
```

## Swept for the same bug class elsewhere

`go vet` can't check `fmt.Sprintf(e.LoadedStrings.X, ...)` because the format string isn't constant — which is how this class slips through, and it has bitten twice now. Scanned all 86 `strings.json` values passed to Sprintf against their shipped defaults: **no arity mismatches**. `badUDRatio` and friends contain a literal `%` but are never formatted, so they're fine.

## Docs

`docs/sysop/advanced/string-editor.md` covered `|XX` placeholders but never mentioned that some strings carry format verbs whose count and order must be preserved — the thing most likely to break a string edited in the TUI. Added a short section.

## Summary by CodeRabbit

- **New Features**
  - Version banners now support a configurable `%s` placeholder for displaying the running version.
  - Empty or outdated default templates automatically fall back to the current format, while customized banners are preserved.
  - Invalid templates remain visible without exposing formatting-system error messages.

- **Documentation**
  - Added guidance for supported format verbs, ordering requirements, escaped percent signs, formatting failures, and automatic version-string repair.

### DIFF
--- a/docs/sysop/advanced/string-editor.md
+++ b/docs/sysop/advanced/string-editor.md
@@ -148,7 +148,7 @@ different from `|XX` placeholder codes: a placeholder can be moved or removed fr
 verb must stay, and the **number and order of verbs must not change**. The editor's description line
 names the verbs a string expects, for example:
 
-```
+```text
 Exec: Version String    Version string format (%s=version)
 ```
 
--- a/docs/sysop/advanced/string-editor.md
+++ b/docs/sysop/advanced/string-editor.md
@@ -141,6 +141,24 @@ Runtime placeholder codes like `|MN` (menu name), `|TL` (time left), `|CB` (curr
 
 Keys prefixed with `_` (e.g., `_extra3`) are reserved placeholders and cannot be edited.
 
+### Format verbs
+
+Some strings carry `%s` / `%d` **format verbs**, which the BBS fills in at runtime. These are
+different from `|XX` placeholder codes: a placeholder can be moved or removed freely, but a format
+verb must stay, and the **number and order of verbs must not change**. The editor's description line
+names the verbs a string expects, for example:
+
+```
+Exec: Version String    Version string format (%s=version)
+```
+
+Removing or adding a verb does not fail — it prints `%!d(MISSING)` or `%!(EXTRA int=3)` into the
+middle of the message the user sees. If you want a literal percent sign in one of these strings,
+write it as `%%`.
+
+`execVersionString` is the exception that repairs itself: with no `%s` it prints unchanged and logs
+why, rather than showing a mangled banner.
+
 When saving, the editor writes keys in sorted order and omits internal `_`-prefixed keys, producing clean deterministic output.
 
 ## String Descriptions
--- a/internal/menu/executor_runnables.go
+++ b/internal/menu/executor_runnables.go
@@ -14,7 +14,6 @@ import (
 	"github.com/ViSiON-3/vision-3-bbs/internal/ansi"
 	"github.com/ViSiON-3/vision-3-bbs/internal/terminalio"
 	"github.com/ViSiON-3/vision-3-bbs/internal/user"
-	"github.com/ViSiON-3/vision-3-bbs/internal/version"
 )
 
 func runPlaceholderCommand(c *cmdCtx, args string) (*user.User, string, error) {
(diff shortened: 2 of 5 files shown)


robbiew:
Filed the general case as #237.

This PR hardens exactly one string. The residual risk is that `./strings` lets a sysop edit any of the **90** values passed to `Sprintf` and validates nothing on save — and 25 of those carry two or more verbs, where order matters as well as count. `go vet`'s printf analysis needs a constant format string, so `fmt.Sprintf(e.LoadedStrings.X, ...)` is invisible to it; there is no check at edit time, load time or render time.

The issue records that this has now happened twice in opposite directions: `matrixAccountNotValidated` went from one verb to three against existing one-verb configs (caught pre-merge only because someone was looking), and `execVersionString` shipped with zero verbs where the code expected one — which is this PR. Two occurrences in a short period, in opposite directions, is what makes it worth a mechanism rather than a third one-off.

Suggested there: validate verb signatures at load and log, plus a test asserting each shipped default matches its call site — the second closes the source that produced both real occurrences. Noted that `formatVerbCounts` from this PR is the obvious thing to generalise, since it already handles `%%` and the `100% Go` case where the verb is `%<space>`.

robbiew:
Addressed in the latest commit.

You're right, and the mismatch was with my own comment: it claimed "carrying one unchanged means no sysop ever edited it", while the code trimmed before comparing — which is a looser test than that sentence describes.

It matters more than it might look, because whitespace is not incidental in a banner: leading spaces indent it on screen, so `"  <default>"` is a plausible deliberate edit rather than stray formatting. The comparison is exact now.

The blank-value fallback stays, since a banner of nothing but spaces prints an empty line rather than a version — but that is a different question from whether a value matches a default, and the two checks are now documented as separate rather than sharing a trim.

Worth naming the trade-off this leaves: a sysop keeping an indented copy of the old banner keeps the stale version baked into it. They are not left guessing, though — it has no `%s`, so the warning added in this PR names the problem and what to add. Silently overwriting their edit seemed the worse of the two.

`TestSupersededMatchIsExactNotTrimmed` pins it, and against the trimming version reports:

```
an indented banner was overwritten:
  had  "  |15ViSiON/3 Go Edition - v0.1.0 (Pre-Alpha)|07"
  got  "|15ViSiON/3 Go Edition - %s|07"
```

robbiew:
@coderabbitai review

robbiew:
Fixed — the fence is `text` now, since it holds a terminal excerpt. Checked the rest of the file while there; no other unlabelled fences remain.

Issue #242, opened 2026-09-08
./ue TUI should align with base UI settings with ./config and ./strings and use arrow keys to move between fields


robbiew:
1. use ./config's resize handling, loading, background colors, headers, footers, etc.
2. right now you have to scroll vertically to get to column 2 - you should be able to use arrow keys to do this

PR #254, opened 2026-09-08
fix(menuedit): adopt the shared TUI chrome and fix its geometry (#252)

### LINKED ISSUE #252: ./menuedit TUI should follow same base layout/behavior as ./config TUI

robbiew:
Closes #252.

First of two PRs from `docs/dev/menu-user-editor-audit-todo.md`, which this PR also adds. `./ue` (#242) follows.

## Why

`./menuedit` was the last editor still rendering from its own ANSI 256-index palette, with no backdrop art and hand-built title and help bars. `internal/tuiart` exists to hold exactly that chrome — it pins the VGA hex values because most terminal themes map ANSI blue to a bright, low-contrast shade that washes out the white-and-yellow-on-blue UI. So `./config` and `./strings` rendered one way and this editor another. It now aliases `tuiart` the way `internal/configeditor` does.

## The part worth reading

Adopting the backdrop meant every screen had to know which row it was painting, and that is what surfaced the real defects. **Three of the four screens did not fill their terminal**, at every size tested:

- The **menu edit** screen emitted one row too many, pushing content off the bottom and scrolling the terminal. The cause was more specific than "magic constants": `menuFields()` returns **13** fields, but the constant `menuEditFields` claimed **7** and the box height was written for that smaller list. The field list grew and the geometry never followed. Both `menuEditFields` and `cmdEditFields` turned out to be dead code — nothing read either — so they are deleted and both edit screens size their box from `len(fields)`.
- The **command edit** screen overflowed the same way.
- The **command list** emitted one row too *few* on any terminal taller than 25, leaving the bottom row unpainted. It was correct only at 80×25.

Two more defects surfaced during the work:

- The **actively-edited field row was one cell too wide** on both edit screens: the padding was computed against `m.textInput.Width`, but `textinput.View()` renders a cursor cell after the text. Both sites now measure the rendered width, which is what `./ue`'s password dialog already does. This one was invisible to the first probe, which never entered the field-edit modes — the inverse of the artifact trap noted below.
- The confirm and input **dialogs rebuilt the row to the right of the dialog as flat fill**, which erases the backdrop from those rows. They now preserve both sides with `padToCol`/`skipToCol`, as `./config` and this editor's own help overlay already did. `skipToCol` was defined here but had no callers.

## Where the shared scaffold went

As `tuiart.Screen`, not by extracting `configeditor.listBox`. `listBox` also owns that editor's box styling, and moving it would put its byte-identical golden captures at risk for no gain. `Screen` carries only what both remaining editors need: row-tracked background fill, and `Split`, which derives both paddings from one declared fixed-row count.

`Split` deliberately does **not** floor either padding at 1. That floor is what turned a miscount into content scrolling off the screen instead of a visible blank row.

## Verification

`view_geometry_test.go` pins 11 modes × 6 terminal sizes (80×25, 100×30, 120×45, 160×60, 200×100, and an undersized 60×15) — every row exactly terminal-width, every screen exactly terminal-tall, matching the guarantee `./config` and `./strings` already carry.

It drives real keystrokes into each mode rather than assigning `m.mode`. Assigning the mode leaves `textInput` at its constructor width and reports width failures that are artifacts of the test; two hits in the first probe were exactly that.

Full suite, `-race` on the touched packages, `go vet`, `gofmt`, and `git diff --check` all clean.

**Still outstanding:** a manual visual pass in a real terminal. The prior audit's manual pass found two defects that every geometry and golden check passed cleanly, so the automated suite does not retire it.

## Summary by CodeRabbit

- **Improvements**
  - Updated the menu editor’s visual styling for a more consistent interface.
  - Improved screen layout across terminal sizes with dynamic spacing and sizing.
  - Corrected field widths and row-count issues in editor views.
  - Dialogs now preserve surrounding background artwork instead of replacing it with a flat fill.
  - Backdrop artwork remains consistent when the terminal is resized.
  - Prevented long input values from overflowing dialog and editor boundaries.

### DIFF
--- a/docs/dev/menu-user-editor-audit-todo.md
+++ b/docs/dev/menu-user-editor-audit-todo.md
@@ -1,6 +1,7 @@
 # Menu editor and user editor TUI audit
 
-Created: 2026-09-08. Status: in progress.
+Created: 2026-09-08. Status: in progress. Section 1 and 2 shipped for
+`./menuedit`; `./ue` outstanding.
 
 Related issues:
 
@@ -46,8 +47,10 @@ Established by static reading plus a throwaway geometry probe modelled on
 `internal/configeditor/view_geometry_test.go`, run over both editors at 80×25,
 100×30, 120×45, 160×60, 200×100 and an undersized 60×15. The probe drove real
 keystrokes into each screen rather than assigning `m.mode` directly; assigning
-the mode leaves `textInput.Width` unset and produces false width failures that
-are artifacts of the probe, not defects in the editors.
+the mode leaves `textInput.Width` unset and produces width failures that are
+artifacts of the probe. Two of the first pass's hits were exactly that. Note the
+inverse trap as well: the first probe never entered the field-edit modes at all,
+and so missed a real one-column overflow there. See the section 2 results.
 
 ### Shared chrome not adopted (both editors)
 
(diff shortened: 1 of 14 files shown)


robbiew:
## Review comments addressed

**Two of the three were real, and I had them half-fixed.** Thanks — the filled-input case is exactly what my geometry test could not see.

### Bounded the active text input (CodeRabbit `view_menu_edit.go:134`, Copilot `dialogs.go:138`)

Both real, same root cause. An earlier commit in this PR changed these sites to *measure* `textinput.View()` rather than assume `textInput.Width` — right, but incomplete. Measuring only corrected the **padding**: when the view came out wider than the space available, the padding fell to zero and the oversized view was still emitted, so the row overran anyway. Bounding the view is what actually holds the row width.

Reproduced before fixing:

```
menu_edit_field  80x25: row 8  is 82 cells wide, want 80   (55-col Prompt Line)
add_menu         80x25: row 13 is 81 cells wide, want 80   (filled 8-col input)
```

`tuiart.FitInput` now does the bound-and-measure in one place, since `./ue` has the same pattern.

**Why the suite missed it:** `TestViewGeometry` drives real keystrokes but types nothing, so every input it renders is empty. `TestViewGeometryWithFilledInputs` fills each one first, and opens the menu editor's **widest** field by name — at 20 columns the first field's view still fits its row with the cursor cell attached, so opening it would have proved nothing. Reverting either fix makes the new test fail with the exact overruns above.

### Not changed: the narrow-terminal clamp (CodeRabbit `view_command_edit.go:35`)

This one does not reproduce. The claim is that a 60-column terminal gets 76-cell box rows and the `60x15` geometry case fails. It cannot: `WindowSizeMsg` clamps `m.width` to `minWidth` (80) before any view runs, so the model is **80×25** there, not 60×15:

```
after WindowSizeMsg{60,15} -> model width=80 height=25
```

That 80-column minimum with clamping is the documented baseline shared by all four editors (`./config`, `./strings`, `./menuedit`, `./ue`), and the `60x15` case passes. Clamping the box widths on top would be dead code guarding a state the model cannot reach.

### Already fixed

`bgFillStyle` — removed in f8e16a4, which is what was failing lint.

PR #255, opened 2026-09-08
fix(ue): adopt the shared TUI chrome and navigate fields by column (#242)

### LINKED ISSUE #242: ./ue TUI should align with base UI settings with ./config and ./strings and use arrow keys to move between fields

robbiew:
Closes #242. Second of two PRs from the menu/user editor audit; #254 covers `./menuedit` (#252).

**Based on #254** — it adds `tuiart.Screen`, which this PR uses. Retarget to `main` once #254 merges.

## Chrome

`./ue` rendered from its own ANSI 256-index palette, with no backdrop art and three hand-built title/help bar styles. It now aliases `internal/tuiart` the way `./config` and `./menuedit` do, so all four binaries finally agree on the VGA palette — the pinned hex values matter because most terminal themes map ANSI blue to a bright, low-contrast shade that washes out the white-and-yellow-on-blue UI. `editTitleStyle` was a near-duplicate of the list screen's bar and is now the same style.

Both screens are rebuilt on `tuiart.Screen` so the background comes from the backdrop at the row being painted. `./ue`'s geometry was already exact and stays exact. Unlike `./menuedit`, every one of its dialog overlays already preserved the screen behind it with `padToCol`/`skipToCol`, so adopting the backdrop needed no work there.

## The part worth reading — navigation

The edit screen has laid its fields out in two columns since it was written: `fieldDef` carries `Col` (3 or 50) and `Row` verbatim from `UE.PAS` `Proc_Entry`. But Up and Down walked the field *slice* linearly and Left and Right did nothing at all. Reaching `Validated` from `Handle` — the field directly beside it — took **eleven presses of Down** through the whole left column. The geometry needed to fix this was already in the data.

- **Up/Down** now move within the field's own column, wrapping at that column's ends.
- **Left/Right** move to the other column, landing on the editable field whose row is nearest. Right from the bottom of the left column clamps to the last editable right-column field rather than jumping to the top.
- **At the outer edges they hold position.** With only two columns, arrow keys read as spatial movement, and teleporting across the screen from an edge would not.

Tab, Enter, `ctrl+home` and `ctrl+end` are untouched — they still advance in field order, which is the existing muscle memory. The arrow keys are purely additive, so no help-bar text changed. `ftDisplay` fields stay unreachable. `columnOf` treats anything that is not `Col == 50` as the left column, so a field added later with an unexpected `Col` still navigates rather than silently becoming unreachable.

## Incidental

- `renderField` now **measures** what `textinput.View()` renders instead of assuming `Width+1`, reconciling it with `overlayPasswordDialog`, which already measured because "textinput.View() width varies". The same assumption was a live one-column overflow in `./menuedit` (#254).
- **`modeSearch` is unreachable** — nothing assigns it, so the search prompt row in `view.go` is dead code, though `updateSearch` exists to handle it. Left in place rather than deleted, since the intent is clearly that it be wired up. The geometry test exercises the row by setting the mode directly so its width math cannot rot unnoticed. Worth its own issue: finish it or remove it.

## Verification

`view_geometry_test.go` covers 19 modes × 6 terminal sizes (80×25, 100×30, 120×45, 160×60, 200×100, undersized 60×15) — every row exactly terminal-width, every screen exactly terminal-tall. `nav_test.go` pins the column behavior: the reported defect, the edges, the wrap, the clamp, and that every editable field stays reachable by arrow keys alone.

Full suite, `-race`, `go vet`, `gofmt`, `git diff --check` all clean. Every model in the tests is built over a `t.TempDir()` fixture; nothing touches the live BBS.

**Still outstanding:** the manual visual pass over both editors in a real terminal. The prior audit's manual pass found two defects that every geometry and golden check passed cleanly, so the automated suite does not retire it. Golden captures are deferred for the same reason — the screens are still settling, and captures taken now would mostly record churn.


robbiew:
## Update: `modeSearch` is finished rather than removed

The PR description flagged `modeSearch` as unreachable dead code deserving its own issue. Looking at it properly, removal was the wrong call: `updateSearch`, the configured `searchInput`, and the prompt row in `view.go` were all **complete and correct**. Nothing anywhere assigned `m.mode = modeSearch`. The only missing piece was an entry point — six lines.

`/` now opens it, the same key `./strings` binds, for the same search-forward-from-the-cursor-and-wrap behavior.

Three things came with it:

- **A miss now says so.** `updateSearch` returned to the list silently when nothing matched, which reads as the key not having worked.
- **The view tests `modeSearch` before `m.message`.** Both draw on the row under the list box, so a flash message left over from a previous action would have hidden the prompt the user just opened.
- **The help overlay's `dialogH` is derived as `len(helpLines)`** instead of a hand-maintained `19`. Adding the line that advertises `/` would otherwise have left the box off-centre — the same defect that had `./menuedit`'s menu edit screen sized for seven fields after its list grew to thirteen. Derived, it cannot drift again.

The geometry test now reaches `modeSearch` by pressing the key rather than assigning the mode, so its six terminal sizes cover the real state. `search_test.go` covers the binding, the match, case-insensitive substring matching, the wrap, the miss, Esc, and that the prompt is not masked by a stale message.

Docs: the sysop guide gains `/` in the user-list key table and a paragraph on the behavior. Its field-editor table was also still describing the old linear Up/Down walk, so it is corrected for the column-aware arrow keys added earlier on this branch.

**Follow-up worth filing:** `./strings` has the same silent-miss gap in its own `updateSearch`. Left alone rather than widening this PR.

robbiew:
## Folded in: the `./strings` silent-miss fix

Rather than leaving it as a follow-up. `./strings`' `updateSearch` is near-identical to `./ue`'s, including the absent not-found branch, so keeping them in sync is worth more than the tidier PR boundary.

Nothing else was needed there. Unlike `./ue`, that view already switches on `m.mode` before `m.message`, so its search prompt was never maskable by a stale flash message.

**One doc correction fell out of it.** The sysop guide described `/` as "search/filter strings by name", which is wrong twice over: it searches an entry's name, **key, and description**, and it **jumps the cursor** to the match rather than filtering the list. The corrected entry also records the wrap, the miss, and that the prompt opens empty each time.

`search_miss_test.go` covers the miss and asserts a hit is still reported as a hit. The existing golden captures were unaffected — none of them capture a failed search.

So this PR now carries: the `tuiart` chrome adoption for `./ue`, column-aware field navigation, `modeSearch` finished, and this. Happy to split the `./strings` commit back out if you would rather it landed separately.

robbiew:
@coderabbitai review

robbiew:
## Both review comments addressed — one was hiding a real inconsistency

### `view_edit.go:195` — byte vs rune label width

Valid. `labelLen` counted bytes while every other width helper in the file (`padRight`, `centerText`) is rune-based. Every label is ASCII today so the two agree, but a mixed basis is exactly how geometry drift gets reintroduced. Now `utf8.RuneCountInString`.

### `user-editor.md:82` — the missing `Down`

This one was more than a doc gap, and the fix is in code.

Adding column-aware navigation changed Up/Down in `modeEdit` to move within the field's column — but left `modeEditField`'s copies on `nextEditableField`, the linear slice walk. **So pressing Down twice from `Handle` landed you in different places depending on whether the input happened to be open**, a mode with no obvious visual difference. I introduced that when I added the arrow keys and did not notice.

Down and Up now confirm and move within the column from inside the input too. Enter and Tab still advance in field order — that's the older muscle memory and deliberately not column-scoped.

`TestArrowsMeanTheSameThingWhileEditing` pins it by computing where the arrow goes from outside the input, then asserting the same landing from inside. `TestTabAdvancesInFieldOrderFromInsideTheInput` pins that Tab is deliberately different.

### The docs

The root cause of the confusion: the **"Within a Field" table only ever listed cursor movement** — it never said how to confirm or cancel an edit at all, which is why dropping `Down` from the field-editor table read as a regression. It now lists Enter/Tab, Up/Down and Esc, and states that an arrow means the same thing in both states.

---

Also note for whoever merges: CodeRabbit's earlier `pass` on this PR was a **rate limit**, not a review ("Review limit reached"), and before that it was skipped entirely because the base branch was not `main`. I've requested an on-demand review so this PR gets the same scrutiny #254 got — that one turned up two live geometry overruns.

PR #260, opened 2026-09-09
test(menuedit,ue): archive golden screen captures, and close the audit

robbiew:
Closes out `docs/dev/menu-user-editor-audit-todo.md`. The manual visual pass over both editors passed — the last open item, and the one thing gating the captures.

Follows #254 and #255. Mirrors #245, which closed out the string editor audit.

## Why now and not earlier

The captures were deferred deliberately. Taken while the screens were still settling they would have recorded churn; taken now they archive a rendering **known to be right**, because a human has just confirmed it.

**16 captures each**, at 80×25 and 120×45, across eight interaction states per editor. Several capture a *filled* input rather than an empty one — empty inputs are exactly what hid two geometry overruns during this audit.

## Two problems `./strings` did not face

`./strings` uses `tuiart.Shaded` and has no per-record state, so its captures were trivially reproducible. These are not:

- **The backdrop art is chosen at random on startup.** An unpinned capture would record whichever of the three embedded screens the run happened to draw, and fail on the next run. Both tests pin `tuiart.Arts()[0]`, which is stable because `Arts` sorts by filename.
- **The user editor's records carry timestamps.** The fixtures set none, so the date fields render as `Never` and a capture taken today matches one taken next year. Verified by regenerating and diffing, not assumed.

Both suites run with `-count=3` and regenerated twice to confirm byte-identical output each time.

## What the captures cannot do

Recorded in the doc, because it matters for whoever changes this area next. They freeze *what* is drawn, so an unintended change surfaces as a reviewable diff — but **a bar that stops one cell late looks exactly as correct in a golden file as it does in a terminal**. The manual pass stays in the checklist rather than being retired by the automation, for the same reason the prior audit kept it: that pass found two defects every geometry and golden check passed cleanly.

## Verification

Full suite, `-race` on both packages, `go vet`, `gofmt`, `golangci-lint`, `git diff --check` — all clean.

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive visual regression coverage for the Menu Editor and User Editor across common screens, workflows, input states, and terminal sizes.
  - Added reference captures covering menu creation, editing, deletion, command management, user profiles, searching, password management, help, and confirmation dialogs.
  - Tests now use deterministic data and rendering conditions to produce repeatable results.

- **Documentation**
  - Documented golden-test setup, verification procedures, and known limitations.


robbiew:
## Fixed — and this is exactly what the captures were added for

Confirmed and real. The help overlay was **50 columns wide and centred, one column inside the 52-column menu list box**, so the list's side borders showed through the dialog:

```
              ┌╔════════════════════════════════════════════════╗┐
              │║           ViSiON/3 Menu Editor Help            ║│
```

**Every geometry test passed over it.** The rows were exactly the right width — they just had the wrong characters in them. That is the precise gap these captures exist to close, and the reason the audit kept a manual pass in the checklist rather than treating the automation as sufficient.

The dialog is now **54 columns**, covering the menu list box outright while still sitting comfortably inside the wider command list (72) and edit (76) boxes, where a visible surround is the intended look:

```
             ╔════════════════════════════════════════════════════╗
             ║             ViSiON/3 Menu Editor Help              ║
```

Both fixtures regenerated.

## Guard added

`TestNoMixedBorderPairs`, in **both** editors, scans every capture for adjacent single-line and double-line box characters (`┌╔`, `║│`, `╝┘`, …). I swept `./ue`'s 16 captures too — no collisions there, since its dialogs are far narrower than its 62 and 78 column boxes — so its copy pins that rather than fixing anything.

The audit doc records the find, because it is the concrete example of what row arithmetic cannot see.

PR #261, opened 2026-09-09
feat(users): tell sysops when new users join (#207)

### LINKED ISSUE #207: add login notice for sysops that new users have joined
A login.json sequence to notify sysops that new users have joined. it should give them quick access to View the user or Validate / Set Level to Regular User.
Consider making this real-time (if the sysop is online, optional flag/setting set in ./config) as well as login sequence as part of login.json

robbiew:
Closes #207. Two halves, matching the issue: a notice in the login sequence for the SysOp who was not around, and a real-time page for the one who is.

## Login sequence — mostly already built, never shipped

`NEWUSERVAL` already counts pending accounts, prompts to review them, and opens the pending-validation queue where each user can be **viewed**, **validated (`G`)**, or have their **level set (`F`)** — exactly what the issue asks for. It was registered but absent from `templates/configs/login.json` and from the code-level default sequence. Both now include it.

**It also had a leak.** The prompt ran *before* any permission check:

```
[non-sysop] "3 new users. Review?" -> Y
            "Access denied."
```

The comment said *"Security level check is handled by login.json sec_level field"* — which made a misconfigured config the only thing between an ordinary caller and the pending count. It now checks SysOp ACS itself and returns silently. `sec_level` still works, and is still how a SysOp removes the item.

## Real-time page

New **`notifySysopNewUser`** setting. On signup completion, every online co-sysop-or-above is queued a page through the same mechanism as a user page, so it lands at their next menu prompt rather than interrupting them. The signing-up node is skipped, as are pre-auth sessions.

**It fires regardless of `autoValidateNewUsers`.** An auto-validated signup leaves nothing to review, but somebody joining is still worth knowing at the time it happens.

**Default is `true`, including on upgrade.** `LoadServerConfig` unmarshals over a defaults struct, so a `config.json` predating the key inherits it rather than getting the bool zero value. Deliberate for a notice that only ever adds a line at a SysOp's own prompt — two tests pin both the inherited default and an explicit `false`.

Exposed as **Notify SysOp** under Security Levels in `./config`; text editable in `./strings` as **New User: SysOp Page**.

## Note for the next person adding a string

The catalog entry is **appended, not inserted**. `StringEntry.Number` is an entry's position in that slice and is shown in the list — `metadata.go` documents that a sysop's note about "string 199" should keep meaning the same string. My first attempt inserted it next to the other `newUser*` entries and renumbered everything after it; the string editor's golden captures caught it immediately, which is a nice second earn for those fixtures.

## Verification

8 tests on the page itself (recipients, the level cut-off, the signing-up node, pre-auth sessions, the config flag, auto-validate independence, empty string, and that nothing here can fail a signup), plus 2 on the config defaults. Full suite, `-race` on the four touched packages, `go vet`, `gofmt`, `golangci-lint`, `git diff --check` — all clean.

Docs: the config reference gains the setting and the reasoning; the login-sequence guide gains a `NEWUSERVAL` section and its `RUN:` entry.

## Summary by CodeRabbit

* **New Features**
  * Online SysOps can receive notifications when a new user completes signup.
  * Added configurable notification settings and a customizable message.
  * Added a SysOp-only `NEWUSERVAL` login command for validating pending users.
  * Added configuration-editor controls for enabling or disabling signup notifications.

* **Documentation**
  * Documented signup notifications and the new user-validation command.
  * Clarified that automatic validation and new-user voting cannot be enabled together.


robbiew:
## All four addressed — one root cause, and the reviewer is right

I described `NEWUSERVAL` as costing a SysOp nothing on a quiet day, then shipped it in a way that cost them a cleared screen and a second of their time on **every** login.

### The screen (3 of the 4 comments)

The login runner clears **before** it dispatches:

```go
if item.ClearScreen {
    terminalio.WriteProcessedBytes(terminal, []byte("\x1b[2J\x1b[H"), outputMode)
}
// ...then the handler runs and decides it has nothing to say
```

So `clear_screen` fires whether or not the command goes on to print anything. I had set it in the shipped `login.json`, the code-level fallback, and the docs example. **Removed from all three** — this is a notice, not a screen.

### The pause

`runNewUserValidation` printed `"No new users to validate..."` and slept a second when the queue was empty. Fine while the command was opt-in; as a **default** login item it is a line of noise and a delay on every quiet day. The empty case is now silent.

That message is not simply lost, which is the part worth checking before deleting it. **`VALIDATEUSER`** is the command for opening the queue deliberately — it reports an empty queue and waits, and it is what the admin menu binds. That is the case where silence *would* be wrong, because the keystroke would look like it had done nothing. No shipped menu binds `NEWUSERVAL`, so nothing regresses.

### Pinned

`TestNewUserValDoesNotClearTheScreen` covers both halves of the shipped contract: neither the default sequence nor the shipped `login.json` sets `clear_screen` for `NEWUSERVAL`, and the shipped entry keeps its `sec_level`.

---

**Note for merging:** CodeRabbit has not reviewed this PR either — it hit its review limit again (`Review limit reached`, 02:25). Its green check is not a review. The on-demand checkbox is on that comment if you want one; the command form does not work while rate-limited.

robbiew:
## Both fixed — and the link one has a wider tail

### `login-sequence.md:153` — site-root link

Correct, and I checked it properly rather than taking it on trust, because the repo has links in *both* forms.

The deciding factor is that `docs/sysop/index.html` does not set `relativePath`, which **defaults to `false`** — so docsify resolves links against the **site root**, not the current file. That makes `configuration/configuration.md` right and `../configuration/...` an escape past the root. Every other cross-directory link in `users/` already uses the root-relative form (`nuv.md:223`, `user-management.md:172`), which is what mine now matches.

### `config_test.go` — `t.Skipf` → `t.Fatalf`

Correct. The template is committed, so an unreadable fixture means something is wrong; skipping would have quietly retired both assertions under it rather than reporting the problem.

---

## Follow-up worth filing: 13 pre-existing broken doc links

The same root cause exists in **10 other files**, none touched by this PR:

```
docs/sysop/advanced/event-scheduler.md      docs/sysop/messages/v3mail.md
docs/sysop/configuration/security.md        docs/sysop/networking/README.md
docs/sysop/messages/ftn-echomail.md         docs/sysop/reference/architecture.md
docs/sysop/messages/message-areas.md        docs/sysop/v3net/chat.md
docs/sysop/messages/README.md               docs/sysop/v3net/message-areas.md
```

13 `](../` links that, by the same reasoning, escape the site root and do not resolve. Mechanical to fix, but unrelated to #207 and it would spread this PR across ten more files — better as its own change. Happy to do it if wanted.

robbiew:
## Doc links fixed here — and there were more than 13

Checking the reported link led me to sweep all **293** internal `.md` links in `docs/sysop`. **38 of them did not resolve**, in two shapes:

**13 parent-relative links** across 10 files — the reported class. `../messages/x.md` is resolved from the root, so the `../` escapes the site instead of climbing a directory.

**25 bare same-directory links**, mostly in `v3net/`. `configuration.md` written inside `v3net/` resolves to the site-root `configuration.md`, which does not exist; the file meant is `v3net/configuration.md`. Same root cause, different shape, and the one I would have missed by fixing only what was reported.

The `_sidebar.md` settles the convention — it has always used full paths from the site root (`configuration/configuration.md`, `getting-started/installation.md`). The body text just was not consistent with it.

All 293 now resolve. Verified by walking every link and checking the target exists from the site root.

## Scope of the change

Link targets only:

```
16 files changed, 38 insertions(+), 38 deletions(-)
```

A 1:1 line swap, and every changed line contains a `](` — zero changed lines without one. Link text and anchors are untouched.

Worth noting the shape of this one: the finding was on a link **I added in this PR**. Checking whether I had made the same mistake elsewhere turned up thirty-seven pre-existing instances of it.

PR #265, opened 2026-09-09
docs: add an upgrade guide

robbiew:
There was no upgrade documentation at all — only installation guides and a "keeping binaries updated" how-to for tracking development. That leaves two things an upgrade silently gets wrong.

## 1. New settings are never merged into your config

Both `setup.sh` and `dev-setup.sh` copy a template **only when the target does not exist**, so a setting added since your version is simply absent from your file. What that means differs per file, and the guide says so rather than leaving it to be discovered:

| File | A missing key means |
| ---- | ------------------- |
| `config.json` | The built-in default applies. Usually nothing to do. |
| `strings.json` | Usually a fallback prints the shipped text. A key with no fallback prints nothing. |
| `login.json` | **Nothing happens.** A login step that is not listed does not run. |

So `login.json` is the one that always needs a decision, and `config.json` is usually informational.

## 2. Menu and artwork fixes may not arrive — and it depends on layout

This is the part that surprised me while writing it:

- **Repo in place** — `menus/` is version-controlled, so a pull updates it, and conflicts if you edited menus in place.
- **`dev-setup.sh` instance** — `menus/` is a **copy**, made once at creation and skipped on every later run. `--symlink` applies only to the *programs*, never to `menus/`. So artwork fixes never arrive on their own.
- **Release bundle** — you own the menu set outright.

The guide covers all three, including how to tell which you have, and gives the instance case both remedies: copy the changed file, or replace `menus/` with a symlink once so it tracks from then on.

## Checked, not assumed

Everything is verified against the source rather than described from memory:

- the copy-if-absent rules in `setup.sh:132` and `dev-setup.sh:141,171`
- the hot-reload list in `config_watcher.go:137-149` — note `login.json` **does** hot-reload, which I had wrong until I looked
- which directories are git-ignored (`configs/` yes, `menus/` no)
- the `--symlink` behaviour, which covers binaries only

The key-comparison snippet was run verbatim from a real BBS directory and reports exactly the three cases the worked example describes:

```
== config.json ==   notifySysopNewUser
== strings.json ==  newUserSysopPage (+12 others, all with fallbacks)
== login.json ==    NEWUSERVAL
```

All 294 internal doc links still resolve. Linked from the sidebar, from `installation.md`'s next steps, and from `keeping-binaries-updated.md`, which is adjacent enough to be mistaken for this.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive guide for upgrading existing installations.
  * Documented backup procedures, repository and instance upgrades, configuration handling, restart requirements, menu and artwork updates, and error reporting.
  * Clarified that new configuration settings are not automatically merged during upgrades.
  * Added links to the upgrading guide from installation, maintenance, and Getting Started documentation.
  * Reordered prerequisites information in the binary maintenance guide for improved clarity.

### DIFF
--- a/docs/sysop/_sidebar.md
+++ b/docs/sysop/_sidebar.md
@@ -4,6 +4,7 @@
 * [Overview](getting-started/README.md)
 * [Installation](getting-started/installation.md)
 * [Docker](getting-started/docker.md)
+* [Upgrading](getting-started/upgrading.md)
 
 * **CONFIGURATION**
 * [Configuration](configuration/configuration.md)
--- a/docs/sysop/getting-started/installation.md
+++ b/docs/sysop/getting-started/installation.md
@@ -203,3 +203,4 @@ This builds all binaries, creates a full BBS directory at the target path, and s
 - Review [File Transfer Protocols](files/file-transfer.md) (sexyz ZModem 8k)
 - Refer to [User Management](users/user-management.md) for managing users
 - Set up the [WFC Sysop Console](how-to-guides/wfc-console.md) to monitor live callers remotely over SSH
+- When a new version lands, follow [Upgrading](getting-started/upgrading.md) — config settings added since your version are not merged into your files automatically
--- a/docs/sysop/getting-started/upgrading.md
+++ b/docs/sysop/getting-started/upgrading.md
@@ -1,8 +1,8 @@
 # Upgrading
 
-Upgrading ViSiON/3 replaces the programs. Your settings in `configs/` and your
-`data/` are never touched by any upgrade path — which is what makes it safe, and
-also what makes two steps necessary:
+Upgrading ViSiON/3 replaces the programs. Followed as written, none of the
+procedures below touch your settings in `configs/` or your `data/` — which is
+what makes upgrading safe, and also what makes two steps necessary:
 
 - Settings added since your version are not applied to your existing config
   files.
@@ -39,10 +39,19 @@ data/         users, messages, files, logs
 menus/        if you have edited any menu, .ANS or .CFG file
 ```
 
-`configs/` and `data/` survive every path below, so those copies are insurance
-rather than necessity. `menus/` is the one genuinely at risk, and only on a
-repo-in-place install, where the menu set is version-controlled and a pull can
-overwrite edits made in place.
+Two ways to lose work, both avoidable:
+
+- **`menus/` on a repo-in-place install.** The menu set is version-controlled,
+  so a pull can overwrite edits you made in place. This one bites even when you
+  do everything else right.
+- **`configs/` and `data/` if you extract a release bundle over your existing
+  directory.** A bundle contains both, so extracting in place replaces your
+  settings and can overwrite your data. The bundle section below says how to
+  avoid this; it is the single most destructive thing you can do while
+  upgrading.
+
+Follow the procedures below and the backups are insurance. Deviate from the
+bundle one and they are the only thing between you and starting over.
 
 ## Repo in place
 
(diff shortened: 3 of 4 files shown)


robbiew:
## Fixed — and wrong in the dangerous direction

Correct catch. The intro said an upgrade does not touch your menus; the repo-in-place section then said, accurately, that a pull **does** update `menus/` and can overwrite edits made in the repo.

The failure mode matters: a sysop with in-repo menu edits reads the intro, concludes menus are safe, skips the backup, and loses them. A doc that is reassuring and wrong is worse than no doc.

The intro now states the actual rule:

> Your settings in `configs/` and your `data/` are never touched by any upgrade path […]
> Your menu set is a different story, and which story depends on how you installed. A repo-in-place upgrade **will** update `menus/`, and can overwrite menu edits you made in the repo. An instance or bundle install never updates it, so artwork fixes do not reach you at all.

The backup section says the same, and now marks `menus/` as the one directory genuinely at risk rather than listing three as if they carried equal weight — `configs/` and `data/` survive every path, so copying those is insurance, not necessity.

robbiew:
## All five addressed — two were substantive

### `config.json` reload scope — I had this wrong

The reload list claimed `config.json` applies on save. It reloads, but `config_watcher.go:274` logs a warning on **every** reload:

> some config.json changes require a full restart — fields: ports, keys, IP limits

So a sysop could change `sshPort`, watch the log say "config.json reloaded", and wonder why nothing happened. The guide now separates the two halves: access levels and new-user settings apply immediately; ports, host keys and IP connection limits keep their startup values until a restart.

### The copy loop hid its own failures

```bash
cp ~/git/vision3/$b . 2>/dev/null     # before
```

A permissions failure leaves the old program in place while the operator believes the upgrade finished — an upgrade that half happened, silently. Worse in a guide than in a script, because it teaches the habit. Now:

```bash
src=~/git/vision3/$b
[[ -f $src ]] || { echo "missing from the repo: $b"; continue; }
cp "$src" . || echo "FAILED to copy: $b"
```

…with a line telling the reader not to silence it. A missing source and a failed copy are reported separately, since they mean different things.

### The two lint nits

Directory-listing fence now labelled `text`. The link I added to `keeping-binaries-updated.md` sat in a second blockquote separated by a blank line (MD028) — the two are now one quote, prerequisites first.

### The menu scoping finding

Already fixed. That commit landed **31 seconds** before the review was posted, so it reviewed the previous revision. The current intro scopes it by install type, which is what the finding asked for.

---

Verified: the documented loop passes `bash -n`, every opening fence is labelled, and all 294 internal doc links resolve.

robbiew:
## Fixed — and this is the third time in one document

Right again, and the same shape as the previous two: an absolute reassurance in the intro, contradicted by a caveat further down that a reader who trusted the opening would never reach.

The intro claimed `configs/` and `data/` are never touched by any upgrade path. The bundle section then warns that extracting over your existing directory replaces both — **exactly the mistake this guide exists to prevent**. Someone could read the opening, conclude their settings were safe regardless, and extract in place.

The claim is now conditional on following the procedures. And "Before you start" no longer reassures at all — it names the two concrete ways to lose work, and says which is worse:

> - **`menus/` on a repo-in-place install.** The menu set is version-controlled, so a pull can overwrite edits you made in place. This one bites even when you do everything else right.
> - **`configs/` and `data/` if you extract a release bundle over your existing directory.** […] it is the single most destructive thing you can do while upgrading.

## The pattern, since it happened three times

Copilot caught the menu version, CodeRabbit caught the menu version again with better scoping advice, and now this one. Each time I opened with a clean absolute and then documented the exception below it. For a guide whose whole purpose is "here is what an upgrade quietly does not do for you", leading with a false reassurance is the specific failure that matters.

I grepped the file for the rest — `never`, `always`, `cannot`, `every path` — and the remaining ones are accurate in context: `menus/` genuinely never updates on an instance install, `git pull` genuinely cannot touch a git-ignored `configs/`, and a template genuinely never overwrites an existing file.

All 294 internal links still resolve.

PR #270, opened 2026-09-09
feat(messages): add a date column to the message list (#257)

### LINKED ISSUE #257: Add "date" column to Message List view, ensure 4-digit message numbers fit on row
Current view:

Image

robbiew:
Closes #257.

## The 4-digit number problem

The number column was `"%3d"`. **fmt's width is a minimum, not a maximum**, so message 1000 rendered four columns in a three-column slot and pushed everything after it one place right:

```
msgnum 375     row width = 79 runes
msgnum 3754    row width = 80 runes   <-- right border outside the frame
msgnum 37541   row width = 81 runes
```

The row grew past the 79-column frame and the right border stepped outside it. On an area with thousands of messages that is *every row on screen*, which is what the screenshot in the issue shows.

The other fields had the **mirror-image bug**: truncated by rune via `truncateString`, but padded by byte via `%-Ns`. A multi-byte sender name padded short and pulled the border inward instead.

## The fix

Every cell is now built to an exact rune width, so a row is 77 columns **by construction** rather than by the inputs happening to behave. A number too wide for its column keeps its leading digits rather than widening the row — the frame holding is worth more than the last digit of a six-figure label, and the reader shows the number in full anyway.

## The date column

From the message's own `DateTime`, formatted `01/02/06` to match the reader's existing date substitution (`message_reader_subst.go:157`). A zero time — what a message with no usable date carries — renders blank rather than as year 1.

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                        Main > BBS Ads - Message List                        │
├─────────────────────────────────────────────────────────────────────────────┤
│    N# Subject                         From             To          Date     │
├─────────────────────────────────────────────────────────────────────────────┤
│  3754 Wrong Number PC/Amiga/Linux ... Wrong Number ... All         09/08/26 │
│N 3753 [ANSI] Anime BBS                Shurato          All         09/07/26 │
│  3752 TheForze [New Files]            opicron          All         09/06/26 │
│P 3748 2o fOr beeRS bbS                paulie420        All         09/02/26 │
│  3745 K4JTT BBS                       Jafo             All         08/30/26 │
├─────────────────────────────────────────────────────────────────────────────┤
│                    Page 1 of 752 [1-5 of 3759 messages]                     │
└─────────────────────────────────────────────────────────────────────────────┘
```

Interior 77: status 1, number 5, subject 31, from 16, to 11, date 8, plus five single-space separators. The last is a trailing space so the date does not butt against the right border, the way the status column keeps the number off the left one.

## Keeping it from drifting again

The **column header is generated from the same constants** as the rows, so the two cannot disagree. `TestMessageListColumnWidthsFillTheFrame` fails if the widths stop summing to the interior — nothing enforces that at compile time, and getting it wrong moves the border on message rows only, so the frame looks torn rather than simply narrow.

## Tests

32 cases asserting a row is exactly 79 columns: 1 through 7 digit message numbers, overlong fields, multi-byte fields, a zero date, both output modes, both highlight states. Plus every line of the drawn screen at 79, and that the date shown is the message's rather than today's.

Full suite, `-race` on `internal/menu`, `go vet`, `gofmt`, `golangci-lint`, `git diff --check` — clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected message-list rows so each message displays its own date.
  - Improved handling of long and multibyte text to prevent misaligned columns.
  - Ensured message numbers remain properly formatted within the available space.

- **Improvements**
  - Standardized message-list column widths and row layout for consistent alignment.
  - Improved highlighted and normal row rendering so screen lines maintain a uniform width.
  - Blank dates now display cleanly when no date is available.

### DIFF
--- a/internal/menu/message_list_render.go
+++ b/internal/menu/message_list_render.go
@@ -2,7 +2,9 @@ package menu
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
+	"time"
 	"unicode/utf8"
 
 	"github.com/ViSiON-3/vision-3-bbs/internal/ansi"
@@ -71,9 +73,15 @@ func drawMessageListScreen(terminal *term.Terminal, state *MessageListState, are
 		return err
 	}
 
-	// Column headers (bright white text, total interior: 77 chars)
-	// Layout: Status(1) + " N#" (3) + "  " (2) + "Subject" + pad (33) + "    " (4) + "From" + pad (17) + "  " (2) + "To" + pad (15) = 77
-	columnHeaders := "|11│|15 N#  Subject                               From               To             |11│|07\r\n"
+	// Column headers, built from the same widths as the rows so the two cannot
+	// drift apart. The status column has no label.
+	columnHeaders := fmt.Sprintf("|11│|15%s%s %s %s %s %s |11│|07\r\n",
+		strings.Repeat(" ", listStatusWidth),
+		ansi.PadLeft("N#", listNumWidth),
+		listCell("Subject", listSubjectWidth),
+		listCell("From", listFromWidth),
+		listCell("To", listToWidth),
+		listCell("Date", listDateWidth))
 	if err := terminalio.WriteProcessedBytes(terminal, ansi.ReplacePipeCodes([]byte(columnHeaders)), outputMode); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


robbiew:
## Both fixed — the CP437 one was real and I nearly mis-tested it

### CP437 truncation order

Correct. `toCP437Safe` emits one byte per rune, and CP437's high bytes are not valid UTF-8, so converting before truncating meant `TruncateRunes` decoded them and rebuilt the string through `[]rune`:

```
converted:        43 61 66 82 20 a5 6f a4      Café Ñoñ…
after listCell:   43 61 66 ef bf bd 20 ef …    Caf<U+FFFD> …
```

Every accented character destroyed — and **the column count stayed correct throughout**, which is exactly why the width tests passed straight over it.

Order is now: truncate and measure in Unicode, convert, then pad with ASCII spaces, which are the same byte either way. Pre-existing rather than introduced here, but it lives in the function this PR rewrites.

### Getting the test right took two attempts, which is worth recording

My first assertion was `strings.ContainsRune(out, '�')` — and it **failed on correct output**. Ranging a Go string over CP437's high bytes yields `RuneError` for each one, so a rune-level check cannot tell a corrupted character from a properly encoded `é`. I traced the pipeline stage by stage before concluding the code was fine and the test was wrong.

It compares bytes now: no `ef bf bd` sequence anywhere, and the single bytes `0x82` (é) and `0xa5` (ñ) present. Reverting the ordering makes it fail, so it is pinned rather than merely passing.

### The doc comment

Also right. It claimed a 77-column result while the function deliberately returns 76 — the status character is written by the caller. It now states the width in terms of the constants rather than a number that contradicted the code.

---

Full suite, `-race` on `internal/menu`, `go vet`, `gofmt`, `golangci-lint` — clean.

Issue #323, opened 2026-09-10
design: adopt a Synchronet-style recycle model for config reload (semaphore trigger + idle-gated apply)


robbiew:
Design proposal coming out of #311. Not a bug — a question of which reload architecture we commit to before we widen coverage.

## Background: how Synchronet does it

Synchronet doesn't hot-swap individual settings. Each server — terminal, FTP, web, mail, services — has its own lifecycle and **recycles**: it tears down and re-initializes from `ctrl/*.cnf` wholesale. A recycle is triggered by a semaphore file, by the sysop from `sbbsctrl`, or on a timer, and the terminal server waits until its client count reaches zero before acting. SCFG touches the semaphore after a save, so the editor and the server never have to agree on anything except "something changed."

> **Verified against the sbbs source** — see the first comment below. The shape holds, but the semaphore lives in `ctrl/` (not `data/`), the trigger is a polled mtime check rather than file events, and there is a **third apply lane** (per-node config snapshot at attach) that supersedes part of the two-lane proposal below. Read that comment alongside this.

## Why it's interesting for us

Recycle dissolves the entire problem class the three open bugs describe:

- No per-field setters → nothing to forget to propagate (#322)
- No concurrent readers at the swap point → no torn reads (#321)
- One signal per save instead of eleven file events to disambiguate (#320)

It is a categorically simpler design than field-by-field hot swapping, and it's the only approach that makes the "hard tier" in #311 — `message_areas.json`, `file_areas.json` — tractable at all.

## Why we can't lift it wholesale

Synchronet recycles because it has discrete servers with clean teardown boundaries. `cmd/vision3/main.go` is one linear ~450-line init with `defer` cleanups threaded through it and package-level globals (`menuExecutor`, `messageMgr`, `fileMgr`, `connectionTracker`, `userMgr`, `sessionRegistry`). There is no subsystem abstraction to recycle. Building one is probably more work than fixing #320–#322 combined, and it would touch every startup path.

### 1. Semaphore trigger instead of fsnotify guessing

Have `v3config`'s `saveAll` (`internal/configeditor/update_save.go:19`) touch a sentinel — say `data/reload.now` — after all eleven writes succeed. The BBS watches for that instead of trying to reconstruct intent from a burst of individual file events.

Strictly better than the debounce fix proposed in #320:

- One signal per save, not eleven
- Fires *after* the save is known-complete, so a partial save can't be picked up
[...]

Issue #311, opened 2026-09-10
Investigate: what other areas can be hot-loaded on config change?


robbiew:
Trying to minimize the scenarios where the BBS must be reloaded when a configuration change is made.

robbiew:
## Investigation results

Traced the full config-load path. Summary: the watcher covers less than it looks like it does, and three bugs gate any expansion of it.

### Current state

**Hot-reloaded today** (`cmd/vision3/config_watcher.go`): `doors.json`, `login.json`, `strings.json`, `theme.json`, `config.json` (partially).

**Already live without any watcher** — worth documenting, since the docs say otherwise:

- Menus — `.MNU`/`.CFG` are re-read on every menu display (`internal/menu/executor_run.go:139` → `LoadMenu`), so menu and art edits are already instant.
- `archivers.json` — `archiver.LoadConfig` is called per operation (`internal/file/manager_paths.go:99`, `internal/ziplab/config.go:119`).
- IP blocklist/allowlist — own fsnotify watcher (`cmd/vision3/main.go:610` `reloadIPLists`).
- `ftn.json` — the binkd mailer reloads it each export cycle and each respawn (`internal/mailer/supervisor.go:92` `reloadFTN`). The watcher's `"ftn.json changed — restart required"` (`config_watcher.go:154`) is stale.

### Three bugs that gate everything else

Split out into their own issues.

**1. The debounce collapses multi-file saves — the big one.**

`config_watcher.go:110-119` stops the pending timer on every event and re-arms it with the *newest* `event.Name`. `v3config`'s `saveAll` (`internal/configeditor/update_save.go:19`) writes eleven files back-to-back: `config.json` → conferences → msg areas → file areas → `doors.json` → `ftn.json` → events → `v3net.json` → `protocols.json` → `archivers.json` → `login.json`. All well inside the 500ms window, so **only `login.json` ever gets reloaded**. A sysop who edits access levels in the TUI and saves gets no hot reload at all — the feature is effectively dead from its main entry point.

**2. The "atomic" setters are racy on the read side.**

[...]

PR #327, opened 2026-09-10
fix(config): sbbs-style config reload — polled mtime signal, tracker propagation, atomic snapshots

### LINKED ISSUE #320: fix(config): watcher debounce drops all but the last file in a multi-file save
Found while investigating #311.

## Problem

The config watcher's debounce keeps a single timer and re-arms it with the newest filename on every event, cancelling the previous one:

`cmd/vision3/config_watcher.go:104-119`

```go
if debounceTimer != nil {
    debounceTimer.Stop()
}
debounceTimer = time.AfterFunc(debounceDuration, func() {
    cw.handleConfigChange(event.Name)
})
```

Only the last file to change within the 500ms window is ever reloaded. Every earlier change in the burst is silently dropped.

## Why this matters

`v3config`'s `saveAll` (`internal/configeditor/update_save.go:19`) writes eleven files back-to-back on a single save:

`config.json` → `conferences.json` → `message_areas.json` → `file_areas.json` → `doors.json` → `ftn.json` → `events.json` → `v3net.json` → `protocols.json` → `archivers.json` → `login.json`

All of those land well inside 500ms, and `login.json` is last. So **the only file the running BBS ever hot-reloads after a config-editor save is `login.json`.** A sysop who changes access levels, board name, or a door definition and saves sees no effect until a restart — even though `config.json` and `doors.json` both have working reload paths.

This makes the existing hot-reload feature effectively dead from its main entry point.

## Repro

1. Start the BBS.
2. Run `v3config`, change `coSysOpLevel`, save.
3. Watch the log: only `login.json reloaded` appears. No `config.json reloaded`.

## Fix

Accumulate dirty basenames instead of overwriting a single name — one timer, a `map[string]bool` of pending files, drained on fire:

```go
pending := map[string]bool{}
// on event:
pending[event.Name] = true
if debounceTimer != nil { debounceTimer.Stop() }
debounceTimer = time.AfterFunc(debounceDuration, func() {
    for name := range drain(&pending) { cw.handleConfigChange(name) }
})
```

The map needs its own guard since it is touched from both the event loop and the timer goroutine.

A test that writes several config files in quick succession and asserts every one of them reloaded would lock this down.

## Related

- #311 (parent investigation)

### LINKED ISSUE #321: fix(menu): hot-reload setters are guarded but ~500 config read sites are not (data race)
Found while investigating #311.

## Problem

`MenuExecutor`'s hot-reload setters take `configMu`, but almost nothing on the read side does. The writes are synchronized; the reads are not. That is a data race, not a safe swap.

`internal/menu/executor.go:109-165`:

```go
func (e *MenuExecutor) SetStrings(strings config.StringsConfig) {
    e.configMu.Lock()
    defer e.configMu.Unlock()
    e.LoadedStrings = strings
}
```

Read-site counts across `internal/` (excluding tests):

| Field | Unguarded direct reads | Guarded getter |
|---|---|---|
| `e.LoadedStrings` | ~451 | none exists |
| `e.ServerCfg` | ~47 | `GetServerConfig()`, 30 uses |
| `e.Theme` | ~14 | none exists |

Every `strings.json`, `theme.json`, or `config.json` reload races against every active session that is mid-read. `StringsConfig` is a large struct, so a torn read is a plausible outcome, not just a theoretical one.

Separately, `cmd/vision3/main.go:1847` reads the shared `serverConfig.BoardName` inside the `messageMgr.OnMessagePosted` callback — a runtime path — while `reloadServerConfig` writes `*cw.serverConfig` under `serverConfigMu` (`cmd/vision3/config_watcher.go:233-238`). That callback never takes the mutex. It fires when V3Net is enabled and a message is posted.

## Why this matters now

#311 is about widening hot-reload coverage. Every config blob added to the watcher multiplies this exposure. The read path should be made sound first.

## Suggested fix

Converting ~450 call sites to guarded getters is not realistic. An `atomic.Pointer` snapshot per config blob is:

```go
type MenuExecutor struct {
    strings atomic.Pointer[config.StringsConfig]
    theme   atomic.Pointer[config.ThemeConfig]
    // ...
}
```

Readers get a pointer to an immutable snapshot; the setter stores a freshly loaded one. `e.LoadedStrings.Foo` becomes `e.Strings().Foo`, which is a mechanical rename rather than a lock-discipline change, and it removes the per-read struct copy that `GetServerConfig()` currently does.

For `main.go:1847`, capture `BoardName` at closure creation or read it through `menuExecutor.GetServerConfig()`.

## Verification

There is no `-race` coverage of a reload concurrent with session reads today. `internal/mailer/reload_test.go` is a good model for what this should look like — it races `reloadFTN` against `exportOnce` under `-race` deliberately.

## Related

- #311 (parent investigation)

### LINKED ISSUE #322: fix(config): config.json reload never reaches the ConnectionTracker (node/IP/lockout/rate-limit settings ignored)
Found while investigating #311.

## Problem

`reloadServerConfig` (`cmd/vision3/config_watcher.go:222-254`) pushes the new `config.json` into three places: the shared `serverConfig` var, `menuExecutor`, and `userMgr` (new user level, auto-validate). It never touches the `ConnectionTracker`.

The tracker captures its limits at construction and never updates them (`cmd/vision3/main.go:1577-1590`):

```go
connectionTracker = NewConnectionTracker(
    serverConfig.MaxNodes,
    serverConfig.MaxConnectionsPerIP,
    serverConfig.MaxFailedLogins,
    serverConfig.LockoutMinutes,
    serverConfig.IPBlocklistPath,
    serverConfig.IPAllowlistPath,
)
connectionTracker.SetConnRateLimit(
    serverConfig.EnableConnRateLimit,
    serverConfig.ConnRateLimitHits,
    serverConfig.ConnRateLimitWindowSeconds,
    serverConfig.ConnRateLimitBanMinutes,
)
```

So these `config.json` fields are silently ignored on reload:

- `maxNodes`
- `maxConnectionsPerIP`
- `maxFailedLogins`
- `lockoutMinutes`
- `ipBlocklistPath` / `ipAllowlistPath`
- `enableConnRateLimit`, `connRateLimitHits`, `connRateLimitWindowSeconds`, `connRateLimitBanMinutes`

The rate-limit group is the most annoying of these: `SetConnRateLimit` (`cmd/vision3/main.go:373`) is a working setter that already exists and is simply never called again after startup.

The watcher's closing warning names only `"ports, keys, IP limits"` as restart-required (`config_watcher.go:252`), which does not make clear that node and connection limits are in that set.

Note this is separate from the IP list *files*, which are watched and reloaded correctly (`reloadIPLists`, `cmd/vision3/main.go:610`). It is the configured *paths* and the numeric limits that go stale.

## Fix

1. Add setters to `ConnectionTracker` for the node/IP/lockout limits and the list paths (path changes should trigger a `reloadIPLists` and re-point the file watcher).
2. Give `ConfigWatcher` a `*ConnectionTracker` reference and call them from `reloadServerConfig`, alongside the existing `SetConnRateLimit` call.
3. Decide the shrink policy for `maxNodes` / `maxConnectionsPerIP`: lowering the limit below the current count should reject *new* connections until usage drops, not drop live sessions. Worth stating explicitly in the docs.
4. Update the restart-required warning to list what genuinely still needs one (listener ports/hosts, SSH host keys, `sshEnabled`/`telnetEnabled`, QWK API listener, logging dir/rolling).

## Related

- #311 (parent investigation)

robbiew:
## Summary

First stage of the config-reload work from #311 / #323: the three bugs that gated everything else, fixed with the sbbs-style architecture out of the gate rather than patched in place. Three commits, one per issue, each independently testable.

### 1. Polled mtime replaces the fsnotify watcher (#320)

The old watcher debounced events by re-arming a single timer with the newest filename, so `v3config`'s eleven-file save reloaded **only `login.json`** — hot reload was effectively dead from its main entry point.

The watcher now polls modification times every 2s (Synchronet's `semfile` model, `DEFAULT_SEM_CHK_FREQ`). Polling is level-triggered: a burst of writes leaves every file with a new timestamp and one poll catches all of them, so the bug class is gone rather than patched. It's also self-healing — a file read mid-write fails to parse, the reload is skipped, and the next poll retries because the writer's final mtime differs from the recorded one.

Two explicit triggers added:

- **`configs/reload.now`** — semaphore file; `v3config` touches it only after *all* writes succeed (the early returns in `saveAll` mean a failed save never signals), and sysops/scripts can `touch` it by hand
- **`SIGHUP`** — reloads instead of terminating; not delivered on Windows

Backward mtime changes count as changes (`!Equal`, not `After`), so restoring a config from a backup still reloads — a case sbbs's `<=` comparison would miss.

### 2. config.json reload reaches the ConnectionTracker (#322)

`reloadServerConfig` updated the executor and user manager but never the tracker: `maxNodes`, `maxConnectionsPerIP`, `maxFailedLogins`, `lockoutMinutes`, the IP list paths, and the rate limiter (whose setter existed and was simply never called again) were all frozen at construction.

`ApplyServerConfig` now pushes all of it. Shrink policy: lowering a limit below current usage refuses new connections until usage drains — it never disconnects anyone online. Clearing an IP list path drops the loaded list instead of filtering forever from the startup copy.

Making the paths settable surfaced two latent bugs, both fixed:

- `StopWatching` closed its channel unconditionally — a panic once anything but shutdown calls it (a path change now restarts the watcher)
- `reloadIPLists` read the paths outside the lock — safe only while they were immutable

### 3. Hot-reloadable config behind atomic snapshots (#321)

The setters took `configMu`; the ~530 read sites did not. Every reload was a data race against every session.

Six configs (`strings`, `theme`, `server`, doors, login sequence, protocols) now live behind `atomic.Pointer` to immutable snapshots. Readers keep whatever snapshot they loaded; a reload stores a fresh one; no reader can observe a half-updated config. The fields are unexported, so the unsafe form is no longer expressible — the accessors are the only way in.

One deliberate asymmetry for reviewers: **`GetServerConfig()` returns a copy** while the other accessors return the snapshot pointer. Callers assign it to a local and some adjust fields before use, which would corrupt shared state through a pointer. Documented at both accessors.

This is also the step toward #323's per-session snapshot lane: sessions can later `Load()` once at login and hold the pointer for their lifetime.

## Verification

- Full suite green under `-race`; builds clean on unix and `GOOS=windows`; `go vet ./...` clean
- **The race regression test was validated against the pre-fix code**: first version detected the old race in ~1/10 runs (the test window closed in microseconds); restructured so the writer's iteration count bounds the test — 10/10 detection on the old code, 0/10 false positives on the new
- **The AST lint in `internal/stringformat` would have silently passed over nothing** after the rename (it matched the old `.LoadedStrings.` selector shape and had no floor). Matcher updated and verified lossless — 132 call sites found before, 132 after — and `TestScanFindsCallSites` adds the missing floor
- New tests: burst reload (the #320 repro), sentinel semantics, backward-mtime, missing/recreated files, live poll-loop end-to-end, tracker limit/rate-limit/path-swap propagation, `StopWatching` idempotence, shrink policy, snapshot isolation, copy semantics

## Docs

`docs/sysop/configuration/configuration.md` rewritten around what actually reloads: table of live vs restart-required settings, the `reload.now` / `SIGHUP` triggers, and the connection-limit shrink policy. (The old text claimed most changes need a restart and that `strings.json` was "loaded fresh on each display" — neither was accurate.)

## Not in this PR

No new files were added to hot-reload coverage (`protocols.json` is now a five-line follow-up since `SetProtocols` exists). #323's remaining lanes — per-session snapshots and idle-gated apply for structural configs — build on this and come separately.

Closes #320. Closes #321. Closes #322. Refs #311, #323.

https://claude.ai/code/session_01H4WPLgdtMyXZ7giCS5Ttiq


robbie-blueocean:
Both findings addressed in 2cd405c:

- **Unlocked path reads**: `startWatching` now snapshots both paths under `ct.mu` — and the review missed that `AppendToBlocklist` had the identical unguarded reads; fixed both.
- **Mutable snapshots**: the map/slice setters clone their argument and the collection accessors clone their result, so no caller can reach the shared snapshot from either end. `GetDoorConfig` still reads the shared map without cloning (returns a value; nothing escapes). Deliberately *not* cloned: `Strings()`/`Theme()` — copying several hundred strings per prompt render is real cost, so their read-only contract remains documentation, now stated explicitly. `TestCollectionSnapshotsAreIsolated` pins the new contract from both directions.

PR #328, opened 2026-09-10
feat(config): hot-reload protocols.json, events.json, and conferences.json

robbiew:
## Summary

Stacked on #327 (base branch is `fix/config-reload`; the diff here shows only the new work). Extends hot-reload coverage to the three candidates from #311 that the new watcher and snapshot accessors unblocked: `protocols.json`, `events.json`, and `conferences.json`.

### protocols.json

The trivial one: a watcher target that loads the file and calls `SetProtocols`. Its single read site already goes through the atomic snapshot accessor from #327.

### events.json — `Scheduler.Reload`

Builds and populates a replacement cron before swapping anything, so a config full of bad schedules still leaves every valid one running. Only cron entries are affected:

- **`run_at_startup` events do not re-fire on a save** — they belong to process startup
- Events mid-run finish against the old cron, which drains in the background
- History, the running-event set, and the scheduler context carry over

Making the swap safe meant closing races that were latent while the schedule was set once at boot:

- `Start` scheduled onto `s.cron` directly and iterated `s.config.Events` unlocked; a reload landing mid-boot could race both. `Start` now works on locals snapshotted under the lock, and starts its cron only if a reload hasn't already superseded it — starting the retired instance would fire its events alongside the replacement's.
- `executeEventWithConcurrency` released into `s.concurrencySem` at defer time. `Reload` resizes the semaphore by replacement, and a release into a fresh empty channel **blocks forever**; the semaphore (and ctx) are now captured under the lock, so a release always pairs with the semaphore its slot came from.
- `Start` no longer exits early when there's nothing to schedule — it stays alive so a later reload can add events, and so shutdown still stops whatever cron a reload started.

The scheduler is created after the watcher, so main hands it over via `SetScheduler`; until then an `events.json` change logs that a restart is needed (a boot-time window, not the permanent state it used to be).

### conferences.json — `ConferenceManager.Reload`

The manager was nearly reload-shaped already: `loadConferences` reads outside the lock and swaps both maps under it, and every menu call site guards its lookups because the manager itself can be nil (conferences are optional). A caller whose current conference disappears mid-session already degrades to "None" and re-selects — no new fallback path was needed.

One edge fixed for reload semantics: a zero-byte file returned early **without clearing the maps** — harmless at boot (they were already empty), but on reload it would have kept serving conferences the sysop had just emptied out.

A malformed save never wipes the running definitions: the maps are only rebuilt after a successful parse.

## Verification

- Full suite green under `-race` (scheduler tests run at `-count=8` to shake out flakes); builds clean on unix and `GOOS=windows`; `go vet` clean
- New tests: cron swap on reload, partial-error keeps valid schedules, startup events don't re-fire, semaphore resize with in-flight release pairing, `Reload` overlapping `Start`'s boot scheduling, conference replace/bad-file/empty-file/reader-race, watcher end-to-end for all three files, nil-scheduler and nil-manager guards
- One test-infrastructure fix worth noting: the scheduler tests initially leaked their `Start` goroutine past test end, where its shutdown history-write raced `t.TempDir` cleanup — a flake that only showed under `-count>1`. `startScheduler` now joins the goroutine in `t.Cleanup`.

## Still restart-only

`message_areas.json`, `file_areas.json`, `v3net.json`, and the listener/logging settings — per the docs table this PR updates. The first two are gated on #323's idle-gated apply lane (and the `file/manager.go` lock-ordering fix); `v3net.json` means tearing down live leaf connections.

Refs #311, #323. Stacked on #327.

https://claude.ai/code/session_013hb7nz7xAjkqrPvVK4mLuN


robbie-blueocean:
Both findings addressed in 9fbbcc3: `testEvent` resolves its command through the existing `lookPath` helper (skip, not fail, where `true` doesn't exist), and `ScheduledCount` takes `RLock`. One knock-on: `lookPath` skips via `t.Skipf`, which must not run inside a goroutine, so the race test now builds its events on the test goroutine before spawning.

PR #329, opened 2026-09-10
feat(logging): apply config.json log level changes live

robbiew:
## Summary

Third slice of the #311/#323 config-reload stack: the log level in `config.json` (`logging.level`) now applies to a running BBS on save, instead of needing the restart that usually destroys the state you were trying to debug.

- `internal/logging`: the handler's level goes through a `slog.LevelVar`; new `SetLevel` validates the name exactly as `Init` does and leaves the level unchanged on unknown input. `SetLevel` before `Init` errors rather than silently no-opping.
- `cmd/vision3/config_watcher.go`: `reloadServerConfig` applies the level, normalizing a local copy first the way `Init` does — an absent `logging` block means the default level, not a warning on every reload. Level changes are logged (`from` → `to`).
- Only the *level* is hot. The log directory and rolling settings belong to the writer built at startup and still require a restart; the watcher's restart note and the sysop docs now say precisely that.

## Verification

- `TestSetLevelChangesEffectiveLevel` proves the point of the feature end to end: records below the level are suppressed, then emitted after `SetLevel("DEBUG")`, then suppressed again at `ERROR` — asserted against the actual JSON log output, not the stored value.
- Unknown names are rejected with the current level intact; pre-`Init` calls error; `Init` seeds the reported level.
- Full suite green with `-race`, builds clean on unix and `GOOS=windows`, `go vet` clean.

## Stack

#327 (`fix/config-reload`) → #328 (`feat/config-hot-reload-coverage`) → this. Merge in that order, retargeting each as its base lands. Coordinated with the parallel session working in this repo — it holds `internal/file` (lock-order fix) and is staying out of `internal/logging` and the watcher.

Refs #311, #323.

https://claude.ai/code/session_01H4WPLgdtMyXZ7giCS5Ttiq


robbie-blueocean:
Finding addressed in 2d0ebf8 — the change announcement now goes out at the new threshold or INFO, whichever is higher, so it survives the change it announces. Note: the session that authored this PR has ended; this fix was applied by the other Claude session working the repo rather than leaving the finding orphaned.

PR #330, opened 2026-09-10
fix(file): resolve the muAreas/muFiles ABBA inversion — never hold both locks

robbiew:
## Summary

Resolves the ABBA lock-ordering inversion in `internal/file/manager.go` that its own doc comment flagged as the blocker for any runtime reload of file areas. Based on `main`, independent of the #327 → #328 → #329 stack — it touches nothing those PRs touch.

## The problem

`FileManager` has two mutexes: `muAreas` (areas/tags) and `muFiles` (records). They were nested in both orders:

- `GetFilePath`, `DeleteFileRecord`, `MoveFileRecord`: held `muFiles`, then took `muAreas`
- `loadAllFileRecords`: held `muAreas`, then took `muFiles`

Safe today only because nothing write-locks `muAreas` after construction. The moment a reload exists, the deadlock is real, and it takes three parties — subtle enough to be worth spelling out:

1. a reader holds `muFiles`, requests `muAreas.RLock` — **queued behind**
2. a *pending* `muAreas.Lock` (Go's RWMutex blocks new readers when a writer waits), which waits on
3. a held `muAreas.RLock` whose owner is blocked wanting `muFiles`

## The fix

No ordering discipline to remember — the nesting is gone entirely: **no goroutine ever holds both locks.** Cross-domain operations copy what they need under one lock, release, take the other. The price is the combined view isn't atomic, so:

- mutators **re-search by ID under the write lock** before touching anything (the first search's index can go stale), treating "no longer there" as not-found
- disk I/O in Delete/Move stays under the write lock as before, so same-file operations keep serializing

The doc comment now states the invariant and names the regression test that guards it.

## Verification

The methodology matters more than the numbers here:

- `TestNoDeadlockUnderConcurrentAreaWriteLock` reproduces the exact three-party cycle: two goroutines running the future reload's lock sequence (`loadAreas` + `loadAllFileRecords`) against concurrent `GetFilePath`/`Move` traffic, writer-bounded with a watchdog so a hang fails fast instead of wedging CI.
- **Validated against the pre-fix code**: dropped into a worktree at `origin/main`, it deadlocks 3/3 runs. On the fixed code: 0/5 full-package `-race` failures.
- `TestDeleteRevalidatesUnderWriteLock` and `TestConcurrentDeleteAndMoveSameRecord` cover the new re-validate semantics — the latter asserts a hammered record ends in exactly one state (one area, or gone).
- Full suite green; unix + `GOOS=windows` builds; vet clean.

## Scope

Lock-layer prerequisite **only**. A public `file_areas.json` reload still needs the data-layer work (refreshing `fileTags`, dropping records of removed areas, path changes) under #323's idle-gated apply design — that's lane 3, deliberately not started here.

Refs #311, #323.

https://claude.ai/code/session_013hb7nz7xAjkqrPvVK4mLuN

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of file and area operations during concurrent reloads, reads, moves, and deletions.
  * File paths now reflect the latest record name when files are moved or deleted.
  * Reduced the risk of stale metadata causing incorrect file operations or inconsistent path results.
  * Improved handling of area changes while records are being loaded.

* **Tests**
  * Added regression coverage for file deletion and renaming during concurrent changes.

### DIFF
--- a/internal/file/manager.go
+++ b/internal/file/manager.go
@@ -30,7 +30,12 @@ import (
 // paths read this way can in principle go stale against a future area
 // reload; such a reload is expected to run only with no sessions active
 // (see issue #323), which is also what keeps the rest of a reload's
-// consequences — removed areas' records, changed paths — tractable.
+// consequences — removed areas' records, changed paths — tractable. The
+// same constraint covers in-flight mutators: DeleteFileRecord and
+// MoveFileRecord release muFiles between their in-memory update and
+// saveFileRecords, so a reload landing in that window would re-read
+// metadata.json — still holding the pre-mutation list — and revert the
+// update after the disk operation has already run.
 type FileManager struct {
 	basePath    string               // Base directory for all file areas (e.g., "data/files")
 	configPath  string               // Path to file_areas.json
--- a/internal/file/manager.go
+++ b/internal/file/manager.go
@@ -14,26 +14,23 @@ import (
 // FileManager manages file areas and their associated file records.
 //
 // Locking: FileManager has two mutexes, muAreas (guards fileAreas/fileTags)
-// and muFiles (guards fileRecords). The intended acquisition order is
-// muAreas before muFiles. GetFilePath and DeleteFileRecord invert that
-// order: they take muFiles first and, while still holding it, take muAreas
-// (nested, not sequential — both are held at once). MoveFileRecord does
-// the same nested muFiles-then-muAreas acquisition internally (after an
-// initial, unrelated sequential muAreas check made before muFiles is ever
-// taken). loadAllFileRecords is the only function that nests them in the
-// intended muAreas-then-muFiles order — and it, too, runs only from
-// NewFileManager, so at runtime the nested order is in fact uniformly
-// muFiles-then-muAreas. Both halves of the inversion are therefore
-// confined to construction.
+// and muFiles (guards fileRecords). The invariant is that no goroutine ever
+// holds both at once — there is no acquisition order to get wrong, and a
+// runtime reload of file areas (an exclusive muAreas.Lock() after startup)
+// cannot deadlock against record operations. Historically these locks were
+// nested in both orders (an ABBA inversion, safe only because muAreas was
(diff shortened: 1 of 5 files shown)


robbiew:
Addressed the stale-filename finding in aa0130f — it was real, and worth the "closer look" flag.

**The hazard:** `DeleteFileRecord`/`MoveFileRecord` captured `Filename` in their read phase; `UpdateFileRecord` can rename the record in the unlocked gap, so the disk delete/rename targeted a name another file may have taken over. The pre-split code was immune only because everything ran under one continuous lock hold — this was a genuine regression introduced by the restructure.

**The fix:** both operations now take the filename from the record as it exists under the write lock, after the by-ID re-search. No stale copy exists to act on.

**Verification:** `TestDeleteAfterConcurrentRenameTargetsCurrentFile` forces the exact interleave deterministically (the test holds `muAreas` exclusively, which parks the delete between its phases; the rename and a bystander file under the old name land while it's parked). Run against the parent commit it fails both assertions — bystander deleted, renamed file left behind — and passes on this one.

An honest footnote: the first draft of that test deadlocked itself by calling `UpdateFileRecord` while holding `muAreas` — that method's save path takes `muAreas.RLock`. As live a demonstration of why this package must never nest these locks as any comment could give.

robbie-blueocean:
CodeRabbit's sleep-synchronization nitpick addressed in 7cb85c9 — it was right that the sleep proved nothing about the delete having passed its read phase; a late goroutine spawn would have let the rename land first and the test silently stop discriminating.

The fix uses the locks themselves as the synchronization: the test starts with `muFiles` held so the delete parks at its phase-1 `RLock`, then releases and re-acquires `muFiles` — the write lock is granted only after phase 1's `RUnlock`, so everything after that line provably happens-after the read phase. The residual sleep only covers the goroutine reaching the `RLock` call, and a miss there degrades to a non-discriminating pass rather than a flake.

Re-validated in both directions: 3/3 assertion failures on the pre-fix parent commit, 10/10 clean under `-race` on the fix.

PR #331, opened 2026-09-10
feat(config): idle-gated deferred reload lane — file_areas.json and message_areas.json hot-reload at idle

robbiew:
## Summary

Lane 3 of #323: the **idle-gated deferred reload lane**, with `file_areas.json` as its first target — the piece #330's lock work existed to unblock. Based on `main`, everything else from the roadmap already merged.

## How it works

Structural configs — those whose reload can pull state out from under a live session — get a third treatment, between "hot" and "restart":

1. **Validate at signal time.** A changed file is parsed the moment the watcher sees it. A file that doesn't parse is reported immediately and never queued; the running config stays as it was, and fixing the file re-queues it via its fresh timestamp. The sysop hears about a bad save at save time, not hours later when the board happens to empty.
2. **Queue until idle.** The reload applies on the first poll tick with zero active sessions. Sessions register at connect (before login), so `ActiveCount() == 0` genuinely means nobody is anywhere in the system — including the matrix screen.
3. **Two escape hatches, deliberately different in power:**
   - `configs/reload.now` (touched **automatically** by `v3config` on every save) *queues* structural changes but never applies them over live callers — an automatic save must not bypass the gate. `TestSaveSentinelDoesNotBypassIdleGate` pins exactly this.
   - `configs/reload.force` (touched by **nobody** automatically) is the explicit "I know callers are online, do it now": applies queued structural changes immediately and re-reads everything else besides.

## file_areas.json

`FileManager.Reload` is `loadAreas` + `loadAllFileRecords`: both lookup maps rebuilt, removed areas' records dropped, added areas' records read from their `metadata.json`. A malformed file keeps the old definitions (`loadAreas` parses before mutating). The method documents the no-active-sessions constraint from the type comment — the idle gate is what discharges it. The #330 lock-order regression test now exercises the real `Reload` instead of its stand-in helper.

## Verification

- New tests: queue-while-online → apply-at-idle (within one poll of the last logoff), immediate application on an idle board, validation failure not queued + re-queue on fix, force sentinel bypassing the gate, save sentinel *not* bypassing it, and an end-to-end run through the real `FileManager` (new area invisible while a caller is online, visible after logoff)
- `FileManager.Reload` tests: area add/remove with records following, malformed file keeps old, externally edited `metadata.json` picked up
- Full suite green under `-race`; unix + `GOOS=windows` builds; vet clean

## Left for follow-ups

- `message_areas.json` — same lane, but adds cached JAM base handles and the in-process mailer to the picture
- WFC surfacing of the queue — `PendingReloads()` exists and is tested; the console display is its own change
- #323 lane 2 (per-session snapshots) and the `v3net.json` restructuring

Refs #311, #323.

https://claude.ai/code/session_01X8bsBNQ533XxwP1Z2ph5Lr

## Summary by CodeRabbit

* **New Features**
  * File-area and message-area configuration changes now reload automatically when the board is idle.
  * Changes are validated and queued while sessions are active, then applied when the board becomes idle.
  * Administrators can force pending configuration changes immediately using `configs/reload.force`.
  * Reloaded areas and metadata reflect external updates, removals, and empty configurations.
  * The interface displays pending configuration reloads.

* **Bug Fixes**
  * Existing area definitions are preserved when configuration files are missing or invalid.

* **Documentation**
  * Updated configuration guidance for automatic, deferred, validated, and forced reloads.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,8 +64,9 @@ internal/configeditor/reload.now
 # SSH host keys (generated during setup, not tracked in git)
 /configs/ssh_host_*
 
-# Config reload semaphore (touched at runtime to signal a config reload)
+# Config reload semaphores (touched at runtime to signal a config reload)
 /configs/reload.now
+/configs/reload.force
 
 # Logs and debug files
 *.log
--- a/cmd/vision3/config_watcher.go
+++ b/cmd/vision3/config_watcher.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"os"
@@ -10,6 +11,7 @@ import (
 	"time"
 
 	"github.com/ViSiON-3/vision-3-bbs/internal/config"
+	"github.com/ViSiON-3/vision-3-bbs/internal/file"
 	"github.com/ViSiON-3/vision-3-bbs/internal/logging"
 	"github.com/ViSiON-3/vision-3-bbs/internal/menu"
 	"github.com/ViSiON-3/vision-3-bbs/internal/scheduler"
@@ -59,14 +61,29 @@ type ConfigWatcher struct {
 	connTracker    *ConnectionTracker
 	scheduler      *scheduler.Scheduler // set via SetScheduler after startup; nil until then
 
-	interval     time.Duration
-	targets      []reloadTarget
-	sentinelPath string
+	interval          time.Duration
+	targets           []reloadTarget
+	deferredTargets   []deferredTarget
+	sentinelPath      string
+	forceSentinelPath string
 
-	mu     sync.Mutex           // guards mtimes and stop
-	mtimes map[string]time.Time // last-seen modification time per polled path
-	stop   chan struct{}
-	done   chan struct{}
+	mu      sync.Mutex           // guards mtimes, pending, and stop
+	mtimes  map[string]time.Time // last-seen modification time per polled path
+	pending map[string]bool      // deferred targets queued for the next idle window
+	stop    chan struct{}
+	done    chan struct{}
+}
+
+// deferredTarget is a watched file whose reload is structural: applying it
+// with callers online can pull state out from under a session, so a detected
+// change is validated immediately (the sysop hears about a bad file at save
+// time, not hours later) but applied only once the session registry reports
+// zero active sessions — or when the force sentinel demands it.
+type deferredTarget struct {
+	name     string // basename, for logging and pending listings
+	path     string
+	validate func() error // parse-only check, run at signal time
+	apply    func() error // the actual reload, run at the idle window
 }
 
 // NewConfigWatcher creates a configuration file watcher and starts polling.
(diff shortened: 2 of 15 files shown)


robbiew:
Two additions since the PR opened, both from its own follow-up list:

**WFC surfacing** (ca7809f) — the snapshot now carries `PendingReloads` and the console header shows `RELOAD PENDING: <files>` while anything is queued, so a deferred save is visibly waiting rather than looking like a save that did nothing. The admin server also moves after the menu executor and config watcher in startup order; its getters always closed over `menuExecutor`, which was assigned later — harmless only because nothing calls them before a connection arrives. With a third closure joining, the ordering is now correct instead of accidentally survivable.

**message_areas.json** (eb5066f) — the last hard-tier file, and it turned out far tamer than the #311 investigation predicted: JAM bases are opened on demand (no handle cache to invalidate), and the non-session writers (QWK API, mailer, V3Net inbound) only post messages — nothing but sysop-session paths ever rewrites the file, so the idle gate covers everything. `MessageManager.Reload` is the area load plus dropping the thread/MSGID index caches (keyed by area ID; stale ones would answer for the wrong base after a BasePath change). Documented limitation: V3Net routing binds at startup and is not rewired — that waits for the v3net restructuring.

With this, every file from #311's table hot-reloads except `v3net.json`. PR body's remaining follow-ups: lane 2, v3net.

robbiew:
Findings addressed in 2c531cc — three accepted with regression tests, three declined with reasoning:

**Accepted:**
- **Pending cleared before apply** (Copilot + CodeRabbit): both right, and my in-code rationale for the old behavior was wrong — it ignored transient failures, which validation can't gate. Pending now clears only on success; failures retry each idle poll, with the recurring error log as deliberate visibility.
- **Stale pending after an invalid follow-up edit** (CodeRabbit): the best catch of this review. Valid edit queues → invalid edit lands before idle → the idle window would have applied the invalid file. Validation failure now clears the mark.
- **Missing `file_areas.json` treated as success** (Copilot): worse than stated — `loadAreas`' fresh-install courtesy of creating an empty file meant a deletion + `reload.force` would eventually **wipe every area** via the created file's fresh timestamp. `Reload` now refuses a missing file outright.

**Declined:**
- **Atomic session-admission gate** (CodeRabbit, heavy lift): the window is real but benign, and now documented at the apply site plus narrowed with a per-apply idleness re-check. The managers' reloads take their write locks, so a just-admitted session's first structural op serializes after the swap, and mutators re-validate by ID under the write lock (#330). A registration-time gate would put a reload lock on every connect to close a microsecond window those semantics already cover.
- **Semantic validation at queue time** (CodeRabbit): skip-with-warning is boot behavior; a file that loads N areas at boot loads N at reload. Divergence between queue-validation and load semantics would be the actual bug. The warnings are logged, not silent.
- **Pre-publish record staging** (CodeRabbit): partial metadata failure has boot parity too, and with the retry fix above a failed reload now self-heals on the next poll rather than stranding.

robbiew:
CodeRabbit's second round addressed in c5d450c — all three accepted; each sharpened a fix from round one rather than opening a new front:

- **Empty `message_areas.json` rejected at validation**: right — `json.Unmarshal` errors on zero-length input, so the exact edit the empty-file support exists for could never queue. Validation now accepts it.
- **Stat-then-load gap in the missing-file refusal**: right — a deletion between the check and the read still reached `loadAreas`' create-and-succeed path. The policy now lives inside `loadAreas`' own read (construction creates, reload errors on the same `os.ReadFile` result), so there is no gap at all.
- **Index reset outside the map-swap critical section**: right, and the subtlest of the three — a lookup between the two locked sections could pair the new area maps with an index built for a previous `BasePath`, and matching count/modcounter would let validation accept it. The reset now shares the swap's critical section.

Base merged forward into #332 (1a3c4f3). This is the level of review round I'd hoped for — three real sharpenings, zero noise.

PR #332, opened 2026-09-10
feat(v3net): live subscription changes — area browser applies instantly, hand edits ride the idle lane

robbiew:
## Summary

`v3net.json` subscription changes now apply live — the last file from #311's table, and the one with the most direct sysop pain: the in-BBS area browser writes the file and then says **"Restart to activate."** It now activates.

Stacked on #331 (base `feat/deferred-config-reload`) because the hand-edit path rides its deferred lane; the diff here shows only the v3net work. Implements the design the earlier session posted on #311.

## The three structural changes

**1. Leaves get individual lifecycles.** Each leaf runs against a child of the service context with its own cancel + done channel, so one leaf can be stopped — including mid-subscribe-backoff or mid-SSE-stream — and *waited for*, without touching the others. Stops are guarded by a per-run `sync.Once`: shutdown, `Close`, and a reload can all reach the same leaf, and latecomers block until the first stop completes. (The first version raced exactly there; `-race` caught it, and the test that hung on it now asserts a stop interrupts a leaf parked in its retry loop.)

**2. The service drops its read-only-after-Start contract.** `leaves`, `leafByNetwork`, and the area bindings move behind a mutex — the same treatment #321 gave the menu executor, at a scale where a mutex is the right tool (a dozen accessors, all in one file). Bindings now carry the origin line, so the tearline and message-forwarding callbacks in `main` read through `svc.AreaBindingFor()` instead of closing over a map built once at boot. `ConfigureLeaves` absorbs `main`'s leaf-building loop: startup and reload share one implementation.

**3. `ReloadLeaves` is a full stop-and-rebuild, not a per-network diff.** Subscription changes are rare sysop actions; rebuilding is the simplest shape that cannot leave a half-old, half-new set behind. Costs stated where they land: unchanged networks re-subscribe (idempotent), and a caller mid-chat on a V3Net leaf loses that stream.

## Two entry points, same split as the rest of the reload system

- **Area browser** (subscribe/unsubscribe): applies **immediately** — explicit sysop action. Messages become "Subscribed to X. Now active." / "Now inactive.", falling back to the restart text if the live apply fails or the service isn't running.
- **Hand edits to `v3net.json`**: watcher-detected, validated at signal time, applied through #331's **idle-gated deferred lane** — because a rebuild cancels chats, and a hand edit carries no "do it now" intent the way a browser action does.

**Still restart-only:** hub settings, the `enabled` flag, keystore/dedup paths.

## Verification

- `TestReloadLeavesSwapsSubscriptions`: old network stopped (with a hang-detector proving a mid-backoff leaf is interruptible), new network running, bindings swapped, sends to the removed network silently no-op per the unconfigured-network contract
- Reload-before-Start, and a reader-race test over every accessor real callers use, under `-race` at `-count=3`
- Full suite green; unix + `GOOS=windows` builds; vet clean
- Docs: `v3net.json` moves to the idle-applied table with its caveats; the "V3Net needs restart" note on message-area reloads updated to point at the rebind path

With this, **every configuration file from #311's original table hot-reloads** — the last restart-only items are listener ports/hosts, SSH keys, protocol enables, the QWK API listener, logging dir/rolling, and V3Net's hub/identity settings, all deliberately so.

Refs #311, #323. Stacked on #331.

https://claude.ai/code/session_01X8bsBNQ533XxwP1Z2ph5Lr

### DIFF
--- a/cmd/vision3/config_watcher.go
+++ b/cmd/vision3/config_watcher.go
@@ -139,6 +139,12 @@ func NewConfigWatcher(rootConfigPath, menuSetPath string, menuExecutor *menu.Men
 			validate: cw.validateMessageAreas,
 			apply:    cw.applyMessageAreas,
 		},
+		{
+			name:     "v3net.json",
+			path:     filepath.Join(rootConfigPath, "v3net.json"),
+			validate: cw.validateV3Net,
+			apply:    cw.applyV3Net,
+		},
 	}
 
 	// Record current timestamps so the first poll does not reload everything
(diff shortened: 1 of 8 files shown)


robbiew:
Findings addressed in 33b3e32 — two accepted, one declined:

**Accepted:** the test fixture's hand-rolled `itoa` (garbage past 99 — now `strconv.Itoa`, as it should have been), and throttling `SendLogon` in the reader-race loop, which spawns a goroutine per leaf per call. The spin itself stays — it's the interleaving pressure that makes the `-race` coverage meaningful.

**Declined:** the `%s`-with-error claim. Go's `fmt` invokes `Error()` for `%s` on an error operand ([fmt docs, 'Except when printed using the verbs %T and %p, special formatting considerations apply for operands that implement certain interfaces' — error is the first listed](https://pkg.go.dev/fmt)); `%!s(...)` appears for nil errors or non-error non-Stringer types, neither of which applies here. The adjacent pre-existing messages in the same switch already use `%s` with errors, so changing only the new lines would create inconsistency to fix a non-bug.

Base merged forward (127f2ff) to pick up the #331 review fixes.

PR #333, opened 2026-09-10
feat(config): hot-reload ftn.json origin lines — the last item from #311's table

robbiew:
## Summary

Closes the one gap a post-completion audit found against #311's ranked list: **`ftn.json` origin lines** (the "medium tier" `networkOrigins` item) never landed. Per-network origin overrides — and `boardName`, their fallback — were snapshotted into the message manager at boot, so editing an origin line or renaming the board didn't reach new messages' origin lines until a restart.

## Changes

- `MessageManager` gains `SetNetworkOrigins` / `SetBoardName` under its existing lock; the post path's origin read (previously unguarded, safe only while the fields were immutable) now takes `RLock`
- The origins-map construction moves from a loop in `main` onto `config.FTNConfig.NetworkOrigins()`, so startup and reload build the same map by definition
- `ftn.json` joins the watcher's **immediate** targets for exactly this one field — the binkd mailer already re-reads the file on its own export/respawn cycle, so nothing else needs a push (the reload handler's comment says so, to head off "why doesn't this reload everything" confusion)
- `config.json` reloads propagate `boardName` to the message manager's fallback
- Docs: the `ftn.json` row now says origins apply on save; the restart-required list gains an honest entry for `boardName`'s other startup copies (WFC header, V3Net identity, QWK packet headers), which the docs previously implied were live

## Verification

- Origin swap, normalization, and clear-to-fallback tests; a setter-vs-post-path race test under `-race`; a watcher end-to-end test (`ftn.json` edit → `OriginTextForNetwork` returns the new origin)
- Full suite green; unix + `GOOS=windows` builds; vet clean

With this, the #311 table is genuinely complete rather than complete-minus-one.

## Summary by CodeRabbit

* **New Features**
  * FTN network-origin text now updates immediately when configuration changes are saved.
  * Message origins support configurable network-specific text, with the board name as a fallback.
  * Server configuration reloads now apply updated board names to newly generated message origins.

* **Documentation**
  * Clarified which FTN settings apply immediately and which require a restart or the next binkd cycle.

* **Tests**
  * Added coverage for configuration reloads, origin fallbacks, normalization, and concurrent updates.

### DIFF
--- a/cmd/vision3/config_watcher.go
+++ b/cmd/vision3/config_watcher.go
@@ -125,6 +125,7 @@ func NewConfigWatcher(rootConfigPath, menuSetPath string, menuExecutor *menu.Men
 		{name: "protocols.json", path: filepath.Join(rootConfigPath, "protocols.json"), reload: cw.reloadProtocols},
 		{name: "events.json", path: filepath.Join(rootConfigPath, "events.json"), reload: cw.reloadEvents},
 		{name: "conferences.json", path: filepath.Join(rootConfigPath, "conferences.json"), reload: cw.reloadConferences},
+		{name: "ftn.json", path: filepath.Join(rootConfigPath, "ftn.json"), reload: cw.reloadFTNOrigins},
 	}
 	cw.deferredTargets = []deferredTarget{
 		{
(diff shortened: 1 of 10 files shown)


robbie-blueocean:
Addressed in dd25b06 — the finding was right, and it's a regression of the exact race #327 fixed: #332's v3net rewiring reintroduced bare `serverConfig.BoardName` reads in the `OnMessagePosted` closure and `V3NetReload` hook, and this PR making the board name hot-reloadable turned the latent race observable. Both now read through `menuExecutor.GetServerConfig()` — which also means a renamed board flows into wire messages and reload-time origin defaults, consistent with the fallback origin being live. The V3Net subscribe identity (`svc.BBSName`) stays the startup copy, as documented.

robbiew:
Final review round addressed in 4154a1b — all six accepted:

- **V3Net stale fallback** (the substantive one): bindings now store the configured origin verbatim — possibly empty, documented on `AreaBinding` — and consumers apply the board-name fallback at use time through the atomic snapshot. A rename now reaches V3Net tearlines and wire messages, matching the docs, and `ConfigureLeaves`/`ReloadLeaves` lost their `defaultOrigin` parameter entirely.
- **Watcher-path test coverage**: the ftn.json target is now driven through real `NewConfigWatcher` construction + `poll()`, via both a bare mtime change and the save sentinel; `reloadServerConfig`'s `SetBoardName` wiring likewise, including malformed-file preservation.
- **Echomail integration test**: posts through `WriteMessageExt` and reads the JAM text back — configured origin, fallback, and live rename all verified in actual message bodies. (It caught a fixture subtlety: the origin line requires the area's `origin_addr`.)
- **Both doc corrections**: the ftn.json row now scopes what the mailer hot-reloads (enable/disable is restart-only), and the boardName list distinguishes the QWK API service from interactive QWK packets.

Full suite green under -race, both platforms build.

robbiew:
Note on the 23:14 Copilot review: it is anchored to `dd25b06` — the re-review was queued before 4154a1b landed, and it reproduces the previous round's six findings verbatim (it even quotes the pre-fix doc line that no longer exists in the diff). All six are already addressed at the current head; see the disposition comment above. No open findings remain.

Issue #334, opened 2026-09-10
Sysop added to lastcall list even when 'no' is selected


ryanfantus:
My lastcaller list keeps growing with sysop calls in spite of the fact that I am choosing not to put the login into the list. Easy to repro - just login and select no when asked, then see the lastcaller list includes the login anyway (on next login).

robbiew:
That's a bug (and quick fix!). Essentially, it was hiding the login from non-sysops -- but YOU still saw it. fixing it so it hides from all.

https://github.com/ViSiON-3/vision-3-bbs/pull/335

robbiew:
Fixed in #335, now merged to `main`.

**What was wrong.** Answering "no" at the prompt did set the invisible flag on the call record, and the record was stored correctly in `data/users/callhistory.json`. The bug was on the display side: the last callers screen only filtered flagged records for viewers *below* CoSysOp level. So the SysOp always saw their own hidden logins, and they piled up with every "no" answer. The filter now applies to every viewer, regardless of access level.

**Existing entries.** No data cleanup is needed for the accumulated "no" logins. The flag has been written since the prompt was introduced (Feb 20), so every one of those records is already marked `"invisible": true` in the history file. They disappear from the list as soon as you run a build that includes the fix.

Two kinds of entries are *not* affected, because they carry no flag:

- Logins where you answered "yes" (or accepted the default) at the prompt.
- Any SysOp logins recorded before Feb 20, when the flag did not exist yet.

If you want those gone too, the only option is to remove them by hand from `data/users/callhistory.json` while the board is stopped.

robbiew:
One more fix I need to make -- show 20 *visible* callers on the screen (it's counting hidden ones now). I'll follow this up.

PR #336, opened 2026-09-11
fix(lastcallers): fill the screen with 20 visible callers

robbiew:
## Summary

Follow-up to #335 / #334.

Hiding invisible logins from every viewer was correct, but it exposed a second problem underneath: `callHistoryLimit` was doing double duty as both the stored history depth and the displayed row count. Hidden logins still occupy a slot in that 20-record ring, so a SysOp who logs in invisibly steadily evicts real callers from the list.

On this board 16 of the 20 stored records were hidden logins, so the last callers screen rendered **4 rows**. Every further hidden login pushed another real caller out permanently.

## Changes

- `internal/user/manager.go`: `callHistoryLimit` 20 → 200. This is now storage depth only. Hidden logins still take a slot, but the ring is deep enough that they cannot starve the screen.
- `internal/menu/executor_lastcallers.go`: new `lastCallersDisplayLimit = 20` caps the rendered rows. `visibleCallRecords` drops hidden records **before** applying the cap — that ordering is the fix — and keeps the most recent 20 in the oldest-first order `LASTCALL.MID` renders.
- `internal/menu/executor_lastcallers_test.go`: tests for the cap window (keeps newest 20, drops oldest) and for the starvation case (125 records, 100 hidden → still a full 20-row screen).

## Not changed

- The on-disk history and the global call counter. Hidden calls are still recorded, and `GetTotalCalls` reads `nextCallNumber` rather than the history length, so the deeper ring does not shift any stat.
- The **`Calls` column**, which is a separate bug worth its own issue. `LASTCALL.TOP` labels it `Calls` but `@CA@` renders `record.CallNumber`, the board-wide call counter — so J0hnny A1pha shows 63 while `TimesCalled` is 114, and Toady shows 68 having called twice. Every other "Calls" display in the BBS (`executor_admin_users.go:272`, `executor_admin_user_editor_render.go:191`, the `|CALLS` placeholder) uses `TimesCalled`. The related `@NU@` new-user marker tests `record.CallNumber <= 1`, so it can only ever fire for the first call in the board's history rather than flagging first-time callers.

## Verification

- `go build ./...` succeeds
- `go vet ./internal/user/ ./internal/menu/` clean
- `go test ./...` full suite passes
- `gofmt` clean

Note: records already evicted under the old 20-record limit are not recoverable. The screen fills toward 20 rows as new calls arrive.

## Summary by CodeRabbit

* **Improvements**
  * The recent callers display now shows up to 20 visible entries, prioritizing the latest calls.
  * Hidden records no longer reduce the number of visible callers shown.
  * Call history now retains up to 200 records, including hidden login entries.

### DIFF
--- a/docs/sysop/menus/menu-system.md
+++ b/docs/sysop/menus/menu-system.md
@@ -839,14 +839,19 @@ Command usage examples:
 ```json
 {
   "KEYS": "LC",
-  "CMD": "RUN:LASTCALLERS 10",
+  "CMD": "RUN:LASTCALLERS 20",
   "ACS": "*",
   "HIDDEN": false
 }
 ```
 
-- `RUN:LASTCALLERS` - Uses default limit of 10 entries
+- `RUN:LASTCALLERS` - Uses default limit of 20 entries
 - `RUN:LASTCALLERS 25` - Shows last 25 entries
+
+The limit counts entries the viewer actually sees: logins hidden by the
+invisible logon prompt are filtered out before it is applied, so they never
+consume a row. Stored history is deeper than the display limit
+(`callHistoryLimit`), so hidden logins cannot push real callers off the screen.
 
 Template files:
 
--- a/docs/sysop/users/user-management.md
+++ b/docs/sysop/users/user-management.md
@@ -620,12 +620,12 @@ Invisible Logon? Y/n
 
 If they choose **Yes**, the session is flagged invisible for its duration. Invisible sessions are:
 
-- **Hidden from Last Callers** — non-CoSysOp users do not see the call in `RUN:LASTCALLERS`. The call is still logged to `callhistory.json` with `"invisible": true`.
+- **Hidden from Last Callers** — the call does not appear in `RUN:LASTCALLERS` for *any* viewer, including the SysOp: the prompt asks whether to list the login, so declining hides it from everyone. The call is still logged to `callhistory.json` with `"invisible": true`, and the global call counter still advances.
 - **Hidden from Who's Online** — invisible sessions are excluded from `RUN:WHOONLINE` listings and the `NODECT` token count for non-CoSysOp viewers.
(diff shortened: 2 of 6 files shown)


robbiew:
Good catch on the dead cap — that finding was correct and the first pass did not actually change what the screen renders.

`runLastCallers` already had a row limit I missed: `callerLimit` defaulted to 10 and re-sliced the list at the end, and the bundled `MAIN.CFG` passes `RUN:LASTCALLERS 10`. So the new 20-row cap never bound, and worse, it would have broken the documented `RUN:LASTCALLERS 25` contract by making any request above 20 impossible.

Reworked in d0238ce:

- **Dropped the cap.** The pre-existing caller-controlled limit is now the only row limit, so no request is capped. `visibleCallRecords` is a pure filter again.
- **Default 10 → 20**, and `MAIN.CFG` passes 20, so the screen actually fills. An explicit argument still overrides in either direction.
- **Row limiting moved into `mostRecentCallRecords`** so the window is unit-testable, with a test asserting `RUN:LASTCALLERS 25` returns 25.
- Filtering still runs before the limit — that ordering is what keeps hidden logins from consuming rows.
- Docs: corrected the default, and fixed a `user-management.md` line still describing the pre-#334 behaviour where only non-CoSysOp viewers had invisible calls hidden.

On the second point — a fixed 200-record ring can still be exhausted by enough hidden logins — that is true, but at that depth it is the correct outcome rather than starvation: if all of the last 200 calls were hidden, there genuinely were no visible callers to show. The bug was that a 20-record ring was so shallow that a handful of hidden logins evicted recent real callers. 200 gives 10x headroom over the 20-row window while keeping hidden calls on disk, as #335 intended. Guaranteeing a full window regardless of depth would mean trimming the ring by visible count, which seemed like the wrong trade for the benefit.

PR #350, opened 2026-09-12
fix(scheduler): stop an event's orphaned child from holding the run open

robbiew:
Found while working on #345: a scheduler test that ran `sh -c 'sleep 30'` hung on the Linux runner, because the cancellation killed the shell and the forked sleep kept the executor's output pipes open.

The underlying bug is in the executor. It captures stdout and stderr through pipes, so any event that leaves a background child behind (a script that starts a daemon, `sleep 30 &`) blocks `cmd.Run` until that child exits too, however long that is. A timeout kill does not help because it reaches only the process the executor started. The event holds its concurrency slot and running-event mark for the duration, and a BBS shutdown waits on it.

`cmd.WaitDelay` now closes the pipes five seconds after the process exits. The resulting `ErrWaitDelay` means the process itself exited cleanly, so it is logged as a warning and the event is counted on its own exit status; anything the orphan writes after that is discarded.

The new test hangs past a 20s timeout without the change and completes in about 5s with it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Event execution no longer hangs when orphaned child processes keep output streams open after the main process exits.
  * Events now complete successfully after the output wait period, preserving output produced before the main process exits.

### DIFF
--- a/internal/scheduler/executor.go
+++ b/internal/scheduler/executor.go
@@ -3,6 +3,7 @@ package scheduler
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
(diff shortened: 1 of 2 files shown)


robbiew:
Addressed both points from the Copilot review in f0f618a: the test now resolves sleep with lookPath (a missing sleep skips the test rather than letting it pass vacuously), and the shell prints the orphan's pid so the test kills it instead of leaving it running for the remaining 25s.

PR #371, opened 2026-09-18
test(config): make touch a single change from the watcher's view

### LINKED ISSUE #370: Flaky test: TestLivePollLoopPicksUpABurst sees two reloads per file
`cmd/vision3/config_watcher_test.go:TestLivePollLoopPicksUpABurst` failed on CI during #369, on a branch that touches only `internal/menu/ansi_renderer.go`:

```
--- FAIL: TestLivePollLoopPicksUpABurst (0.16s)
    config_watcher_test.go:253: config.json reloaded 2 times, want exactly 1
    config_watcher_test.go:253: doors.json reloaded 2 times, want exactly 1
    config_watcher_test.go:253: login.json reloaded 2 times, want exactly 1
    config_watcher_test.go:253: strings.json reloaded 2 times, want exactly 1
    config_watcher_test.go:253: theme.json reloaded 2 times, want exactly 1
```

All five files, which points at the shared helper rather than any one target.

## The race

`touch` changes the modification time **twice**:

```go
func touch(t *testing.T, path string, age time.Duration) {
	os.WriteFile(path, []byte(`{"changed":true}`), 0644)  // mtime := now
	ts := time.Now().Add(age)
	os.Chtimes(path, ts, ts)                              // mtime := now + age
}
```

The poll loop compares `info.ModTime()` against `cw.mtimes[path]` every 20ms (`config_watcher.go:487-489`). A poll that lands between the `WriteFile` and the `Chtimes` sees change #1 and reloads, then sees change #2 on a later tick and reloads again — two reloads, where the test asserts exactly one.

The test drives all five files from concurrent goroutines, which widens the window, and a loaded CI runner widens it further. Locally it passes 8/8; this is the first failure I have seen across recent runs, so it is rare rather than routine.

Note the assertion is the point of the test — it exists to prove a burst coalesces into a single reload — so relaxing it to `>= 1` would throw away what it is checking.

## Suggested fix

Make `touch` present a single modification to the watcher: write to a temporary file in the same directory, `Chtimes` it there, then `os.Rename` it into place. The rename is one atomic mtime change, so no poll can observe an intermediate state, and the burst-coalescing assertion stays meaningful.

robbiew:
Closes #370.

## Cause

`touch` changed the modification time twice:

```go
os.WriteFile(path, []byte(`{"changed":true}`), 0644)  // mtime := now
ts := time.Now().Add(age)
os.Chtimes(path, ts, ts)                              // mtime := now + age
```

`changed()` (`config_watcher.go:487-490`) compares `info.ModTime()` against the last seen value on every poll, with no debouncing. A poll landing between the two saw one change and reloaded, then saw the other on a later tick and reloaded again — the two-reloads-per-file failure `TestLivePollLoopPicksUpABurst` hit on CI during #369.

## Confirming it rather than assuming it

My first two attempts to reproduce this failed — 15 runs under `-race` and a settle-then-count variant both showed a single reload. The window between the two calls is normally microseconds, so a 20ms poll essentially never lands in it.

Forcing a poll into the window makes it deterministic and shows both the cause and the fix:

```
in-place write + chtimes -> map[config.json:2 doors.json:2 login.json:2 strings.json:2 theme.json:2]
staged write + rename    -> map[config.json:1 doors.json:1 login.json:1 strings.json:1 theme.json:1]
```

Two reloads for every one of the five files, which is exactly the signature CI produced. That all five failed identically is what pointed at the shared helper rather than any single target.

## Fix

Stage the content in a sibling file, set its modification time there, then `rename` it into place. The rename carries the staged time across with the content in one atomic step, so no poll can observe an intermediate state.

The assertion stays `exactly 1`. Proving that a burst of saves coalesces into a single reload is the point of the test, so relaxing it to `>= 1` would discard what it checks.

Every `config_watcher` test uses this helper — `config_watcher_deferred_test.go` calls it a dozen times — so they all stop being exposed to the same race, not just the one that failed.

## Verification

`go test ./cmd/vision3/ -count=3` and the full `go test ./...` are clean, as are `go vet ./...` and `gofmt`.

## Summary by CodeRabbit

* **Tests**
  * Improved configuration watcher test reliability by atomically replacing staged file updates.
  * Added retry handling for concurrent file access on Windows.
  * Preserved timestamp-based test scenarios with explicit temporary-file cleanup and error handling.


robbiew:
Both review points addressed in 9b22821.

The Windows one was a good catch and I had missed the existing machinery entirely: `internal/atomicfile.Replace` already retries exactly the condition `replace_windows.go` documents — Windows refusing to replace a file another handle has open — and the poll loop is stat-ing these paths throughout the test. A bare `os.Rename` would have fixed a Linux timing flake by introducing a Windows one, which is not a trade worth making. It now goes through `atomicfile.Replace`, which is what every other writer in the repo uses.

The duplicated commentary is condensed to a single explanation of the staged write and rename, keeping the original contract line.

